PR #1, opened 2019-10-07
Conference item

MerrettM:


conference-finder/src/App.css
```diff
@@ -20,3 +20,18 @@
 .App-link {
   color: #09d3ac;
 }
+
+.card {
+  width: 100%;
+  box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2);
+  transition: 0.3s
```
dex-mendeley: Forgot the semicolon ;)

conference-finder/src/ConferenceItem.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+const ConferenceItem = ({ conference }) => {
```
dex-mendeley: You could destructure conference here so you won't have to type conference.x, conference.y, conference.z over and over again :) Also, since you're only returning some JSX here without some more logic you could simply return without the curly braces.

So the method declaration would look like this:

const ConferenceItem = ({conferenceName, conferenceTopic, conferenceDateTime, city}) => 
`[ignore the backticks, it's just so github won't render this as html](
        <li>
            <div class="card">
                <h3>
                    {conferenceName}
                </h3>
                <p>
                    {conferenceTopic}
                    {conferenceDateTime}
                    {city}
                </p>
            </div>
        </li>
    )`

conference-finder/src/App.css
RaMorey: What's the transition for? I'm not sure if it's doing anything. Also it's good practice to end all css in a semi colon. Even though it won't break as it is, when you come along to edit the css you might forget that you left it off and wonder why it's erroring

conference-finder/src/App.css
```diff
@@ -20,3 +20,18 @@
...
+}
+
+.card:hover {
+  box-shadow: 0 8px 16px 0 rgba(0, 0, 0, 0.2);
+}
+
+li {
```
RaMorey: If you can, try have a read of the flex stuff we did the other day, it'll be better going forward as your items won't just scale down - the margins will shrink instead

conference-finder/src/ConferenceItem.jsx
j-hutchinson: Alternatively, pass in the full object as a prop then destructure before the return. 
`const {conferenceName, conferenceTopic, conferenceDateTime, city} = conference;`

dex-mendeley: Though in the syntax above you'd still pass in only one object, too. Your solution might be a little bit more readable though :)

conference-finder/src/App.css
```diff
@@ -20,3 +20,27 @@
...
+  margin: 0;
+  padding: 0;
+  list-style-type: none;
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+}
+
+li {
+  width: 33.33%;
```
RaMorey: You should be able to change it to a certain pixel size instead of 33%. It mean's you'll be able to fit more on a bigger screen instead of just spacing them out

conference-finder/src/App.css
MerrettM: To be honest we added it cause it looked nice

conference-finder/src/App.js
```diff
@@ -1,24 +1,18 @@
...
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
+      <ul>
+        <ConferenceItem conference={conferences[0]} />
```
dex-mendeley: If you changed your mockData to a JavaScript file that exports an array you could simply map over that array and return a ConferenceItem each time:

<ul>{conferences.map(conference => <ConferenceItem conference={conference}>))}</ul>

conference-finder/src/App.css
```diff
@@ -20,3 +20,27 @@
...
+.card:hover {
+  box-shadow: 0 8px 16px 0 rgba(0, 0, 0, 0.2);
+}
+
+ul {
+  margin: 0;
+  padding: 0;
+  list-style-type: none;
+  display: flex;
+  flex-wrap: wrap;
```
tonymaynard: If you're looking for the items to be in a column instead of using `flex-wrap` you can use `flex-direction: column;` have a play around with https://the-echoplex.net/flexyboxes/ `flex-wrap` is essentially asking you, do you want to keep everything on each line in the same line or not ( which is what `flex-wrap: wrap;` does ).

conference-finder/src/App.css
tonymaynard: I personally think the way you have done it with percentage is better for your current use case and fixing width to pixels wouldn't be good in this situation but it is down to you guys :)

conference-finder/src/App.js
tonymaynard: Similar to the way we mapped through the book array 2 weeks ago :)

conference-finder/src/App.css
MerrettM: We changed it to px since it keeps the cards a consistent size which we preferred.

conference-finder/src/App.css
MerrettM: Will keep that in mind for when we discuss the visuals with the others.

conference-finder/src/ConferenceItem.jsx
MerrettM: We've now got it passing the full object in and restructuring it on the next line.

conference-finder/src/ConferenceItem.jsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+const ConferenceItem = ({ conference }) => {
+  const { name, topic, dateTime, city } = conference;
+  return (
+    <li>
+      <div className="card">
+        <h3>{name}</h3>
+        <p>{topic}</p>
+        <p>{dateTime}</p>
```
tonymaynard: maybe worth converting this date into a readable format ? by using ` new Date(<dateTime>)` and then taking what you need

conference-finder/package.json
```diff
@@ -5,7 +5,8 @@
   "dependencies": {
     "react": "^16.10.2",
     "react-dom": "^16.10.2",
-    "react-scripts": "3.2.0"
+    "react-scripts": "3.2.0",
+    "styled-components": "^5.0.0-beta.9"
```
tonymaynard: I see you installed 'styled-components' are you going to try and implement them in this PR?

conference-finder/src/ConferenceItem.jsx
MerrettM: Have just done this, definitely an improvement!

conference-finder/package.json
RaMorey: It's fine for now but I would recommend in future not adding them in to the PR until you've started using them :)

conference-finder/src/ConferenceItem.jsx
```diff
@@ -2,12 +2,18 @@ import React from "react";
...
   const { name, topic, dateTime, city } = conference;
+  const date = new Date(dateTime);
   return (
     <li>
       <div className="card">
         <h3>{name}</h3>
         <p>{topic}</p>
-        <p>{dateTime}</p>
+        <p>
+          {date.getDate()}/{date.getMonth() + 1}/{date.getFullYear()}
```
RaMorey: Just a thought, as I've approved already before this commit. But I would recommend you do this logic of building your date on like 6 instead.

I'd also say that using a template literal instead of opening and closing curly braces is nicer to read!

Have a read of these docs, it's a library used to displaying and manipulating dates and is really useful: https://momentjs.com/docs/#/displaying/

PR #3, opened 2019-10-08
Conference list

sophiambolah:
added conferenceList component and code, and updated this in app.js

### DIFF
--- a/src/ConferenceList.jsx
+++ b/src/ConferenceList.jsx
@@ -2,16 +2,12 @@ import React from "react";
 import ConferenceItem from "./ConferenceItem";
 import conferences from "./conferenceDataMock.json";
 
-const ConferenceList = () => {
-  return (
-    <div className="conference list">
-      <ul>
-        {conferences.map(conference => (
-          <ConferenceItem key={conference.id} conference={conference} />
-        ))}
-      </ul>
-    </div>
-  );
-};
+const ConferenceList = () => (
+  <ul className="conferenceList">
+    {conferences.map(conference => (
+      <ConferenceItem key={conference.id} conference={conference} />
+    ))}
+  </ul>
+);
 
 export default ConferenceList;
--- a/src/ConferenceList.jsx
+++ b/src/ConferenceList.jsx
@@ -1,6 +1,6 @@
 import React from "react";
 import ConferenceItem from "./ConferenceItem";
-import { conferences } from "./conferenceDataMock";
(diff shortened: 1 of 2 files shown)


src/ConferenceList.jsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import ConferenceItem from "./ConferenceItem";
+import conferences from "./conferenceDataMock.json";
+
+const ConferenceList = () => {
+  return (
+    <div className="conference list">
```
dex-mendeley: This className property properly works differently to what you expect: You're giving it two classes conference and list. Css classes usually are written in camelCase or with under scores but can't contain whitespaces, so: conferenceList or conference_list

RaMorey: Also to add onto this, I'd argue if you need the div at all? You could have the ul as the parent element and give that the className

dex-mendeley: Yeah, I agree

sophiambolah: Ok we've changed it! :)

src/ConferenceList.jsx
```diff
@@ -0,0 +1,17 @@
...
+      <ul>
+        {conferences.map(conference => (
```
robojeda: It is good practice to try and split your data handling/storing and the presentation of the data.

In this case it's a good idea to pass your conferences array from `App` as a prop  to your `ConferenceList` component. This will make the component more reusable.

MerrettM: so pass import in `App`, pass down to `ConferenceList` and then from there to `ConferenceItem`?

sophiambolah: I don't think we need to reuse the code so for now we'll leave it as is so its easier to read

robojeda: @MerrettM @sophiambolah  Yes, right now the data is being imported from a static document, but if this ends up being an api call you will have to worry about where that logic lives and when you request it. It is good habit to always separate concerns, however if you feel like you it's easier to read like this for the time being go with it.

RaMorey: @robojeda sorry I may have confused them a bit when I went to speak to them earlier and I mentioned too much about reusability and not much about the presentation/data handling. My fault! I agree, once we bring in the api calls and have more logic to perform it will be a lot neater to have it in a separate place, so it's a good idea to do it now!

PR #4, opened 2019-10-08
Styled components

sophiambolah:
We added styled components on conferenceList, conferenceItem and app.js, and removed the app.css file. We also reverted the datetime from using moments back to the .get method.


src/ConferenceItem.jsx
```diff
@@ -1,20 +1,40 @@
 import React from "react";
-import moment from "moment";
+import styled from "styled-components";
+
+const StyledCard = styled.div`
```
tonymaynard: Nice intro to styled components, if you wanted to go a step further you could pass props to the card and then render a different background colour for events that start in the morning or night ( This would only be for your benefit to understand how to render depending on props, and not part of the ticket, but would be useful imo )

package.json
```diff
@@ -3,7 +3,6 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "moment": "^2.24.0",
```
j-hutchinson: 🙌

src/ConferenceItem.jsx
```diff
@@ -1,20 +1,40 @@
...
+const StyledListItem = styled.li`
+  width: 300px;
+  margin: 24px;
+`;
 
 const ConferenceItem = ({ conference }) => {
   const { name, topic, dateTime, city } = conference;
-  const date = moment(dateTime).format("DD/MM/YYYY");
-  const time = moment(dateTime).format("HH:mm");
+  const date = new Date(dateTime);
```
j-hutchinson: I'd suggest for readability creating day, month, year up here and using the variables below in the render. ie 
`const month = date.getMonth();` and so on. It'll make the `p` tag clearer

src/ConferenceList.jsx
```diff
@@ -1,13 +1,23 @@
 import React from "react";
 import ConferenceItem from "./ConferenceItem";
 import conferences from "./conferenceDataMock.json";
+import styled from "styled-components";
+
+const StyledUList = styled.ul`
```
j-hutchinson: Rather than StyledUList, i'd call it something more declarative. Like StyledConferenceList. This will be clearer if you are looking through the DOM in the browser as to what is what

src/ConferenceItem.jsx
MerrettM: We just had a play around with this and got it to work, didn't include it in the latests commit though since it was for practice.

src/ConferenceItem.jsx
MerrettM: Agreed, we implemented this in the latest commit, definitely looks nicer!

src/ConferenceList.jsx
MerrettM: We've also added this to the latest commit.

src/ConferenceItem.jsx
tonymaynard: At this stage theres no harm in including it, as helpers may find a better way to write it as no doubt in the future you will be passing props at some point, similarly it'll help the other guys see how to do it when its their turn to see! but its cool dont worry about it :)

MerrettM: Good point, we've added them in.

src/ConferenceItem.jsx
```diff
@@ -26,10 +26,19 @@ const ConferenceItem = ({ conference }) => {
   const year = date.getFullYear();
   const hours = date.getHours();
   const minutes = date.getMinutes();
+  const inputBackground = ({ hours }) => {
+    if (hours >= 18) {
+      return "#AD9FC8";
+    } else if (hours < 6) {
+      return "#AD9FC8";
+    } else {
```
dex-mendeley: Strictly speaking since this is the default you could leave out the last else and just return :)

MerrettM: We did wonder if the else was needed but I didn't think you could have an else if without an else, thanks we'll fix that.

src/ConferenceList.test.js
```diff
@@ -0,0 +1,38 @@
...
+      id: 2,
+      name: "Royal Parks Half Marathon 2019",
+      dateTime: "2019-10-12T12:34:11Z",
+      city: "Manchester",
+      description:
+        "From Meetingneeds.org.uk: HOW TO BECOME A CHARITY PARTNER AND WHAT’S IN IT FOR YOUR ORGANISATION? As a Charity Partner, you get to be a central part of the Events Industry charity and share in our growth story. Leveraging your engagement with Meeting Needs a",
+      topic: "Sport"
+    }
+  ];
+  describe("given data renders ConferenceList items", () => {
```
j-hutchinson: this describe block isnt necessary, this is the test name `it("given data renders ConferenceList items")`

src/ConferenceList.test.js
```diff
@@ -0,0 +1,38 @@
...
+    it("renders data", () => {
+      let wrapper = shallow(<ConferenceList conferences={mockData} />);
```
j-hutchinson: `expect.assertions(2)` it's good practice to have at the beginning of a test.

src/ConferenceList.test.js
```diff
@@ -0,0 +1,38 @@
...
+      expect(wrapper.find(ConferenceItem).length).toEqual(2);
+      wrapper = shallow(<ConferenceList conferences={[]} />);
```
j-hutchinson: This should be a separate test. One test for when conferences are passed as prop, another test for when its `[]`

src/utils.js
```diff
@@ -0,0 +1,9 @@
+export const getTimestring = date => `${date.getHours()}:${date.getMinutes()}`;
```
j-hutchinson: adding new lines between functions will make this more readable

src/utils.js
```diff
@@ -0,0 +1,9 @@
...
+export const getDatestring = date =>
+  `${date.getDate()}/${date.getMonth() + 1}/${date.getFullYear()}`;
+export const getBackgroundColor = date => {
+  const hours = date.getHours();
+  if (hours >= 18 || hours < 6) {
+    return "#AD9FC8";
+  }
+};
```
j-hutchinson: utils file was a good idea 👍

src/utils.test.js
```diff
@@ -0,0 +1,22 @@
+import { getTimestring, getDatestring, getBackgroundColor } from "./utils";
+
+describe("utils", () => {
+  describe("getTimestring", () => {
+    it("renders correct timestring given a date", () => {
+      const date = new Date("2019-11-12T12:34:11Z");
```
j-hutchinson: I'd add a new line between the setup and expects in a test file. Makes the code more readable. But otherwise good first stab at testing.

src/ConferenceList.test.js
MerrettM: Okay, so taking this and the first change request am I right in saying that we can remove the `describe`, have two `it` statements by themselves, one for each test?

src/ConferenceList.test.js
MerrettM: Does this go before the `it` statements, or inside them? and if it's inside them does that mean when we split this `it` into two we should have one `expect.assertions(1)` in each?

src/utils.js
MerrettM: Thank Dex!

src/ConferenceList.test.js
```diff
@@ -27,12 +27,18 @@ describe("ConferenceList", () => {
...
-      expect(wrapper.find(ConferenceItem).length).toEqual(2);
-      wrapper = shallow(<ConferenceList conferences={[]} />);
-      expect(wrapper.find(ConferenceItem).length).toEqual(0);
-    });
+  it("given list of 2 conferences renders ConferenceList with 2 items", () => {
+    expect.assertions(1);
+
+    let wrapper = shallow(<ConferenceList conferences={mockData} />);
+
+    expect(wrapper.find(ConferenceItem).length).toEqual(2);
```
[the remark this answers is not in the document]
MerrettM: Okay, I think we did those during the testing workshop, when would it be more appropriate to use snapshots over this method?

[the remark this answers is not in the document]
RaMorey: It's a combination of things, and really depends on the way your team want you to do things. Personally, I'd always write a snapshot for the default state of the component. So if you pass it only it's required props or even no props at all I'd have a snapshot to see if it displays correctly by default. 

Then it gets a bit trickier to decide, but for me it just depends if it's something "visual". For example if you had a component that would open a modal when you press a button, I'd have a test that simulates clicking the button and then take a snapshot to see if the modal is open correctly. But you could also do the same test by checking if state has been updated to "modalOpen === true' or something. 

There isn't really a right or wrong most of the time so whatever you feel is more useful for you

PR #5, opened 2019-10-10
GP19-36: Api request

sophiambolah:
made an axios request, implemented loading screen and error handling


src/App.js
```diff
@@ -1,17 +1,42 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import ConferenceList from "./ConferenceList";
+import { getConferenceList } from "./api";
```
RaMorey: This is a very minor comment, but have a think about how you want to structure your imports. It's nice to have an order to them. For example, alphabetical, and then order all of the packages first, then your internal imports, then css

src/App.js
```diff
@@ -1,17 +1,42 @@
...
+    }
+    fetchData();
+  }, []);
+
   return (
     <StyledApp>
-      <ConferenceList conferences={conferences} />
+      {isLoading ? (
+        <p>Loading...</p>
+      ) : (
```
RaMorey: this is just a readability thing for me, so don't feel the need to do it, but I quite like organising ternarys like this:

```
isLoading
? <p>Loading...</>
: <ConferenceList conferences={conferences} />
```

src/App.js
MerrettM: I think I've done this in the latest commit?

MerrettM:
> Does that work without the leading 'http://'?

Ah whoops, no it doesn't!

MerrettM:
Apologies, for that, I clicked the wrong button

MerrettM:
> Does that work without the leading 'http://'?

Ah, no it doesn't, whoops!

src/api.js
```diff
@@ -2,7 +2,7 @@ import axios from "axios";
 
 export const getConferenceList = async () => {
   try {
-    const response = await axios.get("localhost:8080/conferences");
+    const response = await axios.get("http://localhost:8080/conferences");
```
dex-mendeley: For now it's probably alright to keep it there. In the future you probably have some config file with those urls :)

PR #6, opened 2019-10-14
GP19-42: creates reusable button

karenzamani:


src/Button.jsx
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import styled from "styled-components";
+
+const StyledButton = styled.button`
+  text-align: center;
+  background: lightgrey;
+`;
+
+const Button = ({ children, ...props }) => {
```
dex-mendeley: Spreading props is always a little difficult. It's perfectly fine to do so 
but the next developer will not know what props are being passed here. For readability 
it's probably better to name them explicitly

dex-mendeley: Especially since in your case here you're not actually passing any props :)

shanassaf: Agreed it's not needed here when no props are being passed at all, but I'd argue that for html elements with a long list of innate props it's good practice to spread them on.

shanassaf: Developers in the future that add to your code may use your reusable Button component assuming that props are available that you haven't specifically declared.

RaMorey: I'd actually argue that there is a use to passing props in here. While yes they should be naming the things that they are using, there could be other props that they don't necessarily know in advance and don't want to name. For example, disabled etc

RaMorey: Also guys, I'd add some unit tests for your button :) For example, a snapshot of it by itself, a snapshot of it passing in children, maybe even a test to see if your handle click props is triggered when the button itself is clicked.

shanassaf: Yes please, tests in every PR!

RaMorey: And one last one 👀 Can we add the actual functionality of the button explicitly? Like pull the onClick prop out of the props and give it to the styledButton onClick?

dex-mendeley: Snapshot tests are a good idea but testing if a component passes props is a little pointless. That's why we're using React, if we need to test that we would have to test all its functionality

RaMorey: While I would normally agree that testing passing props is pointless, I'd argue that testing that the button renders correctly with it's children is actually testing the functionality of the button. As a result it would help prevent changes that could later be made to the button, like if they delete or remove the children inside.

src/ConferenceItem.jsx
```diff
@@ -31,6 +32,7 @@ const ConferenceItem = ({ conference }) => {
         <p>{getDatestring(date)}</p>
         <p>{getTimestring(date)}</p>
         <p>{city}</p>
+        <Button> Test </Button>
```
RaMorey: I thought I left a comment here but I can't see it... Recommend adding padding to your button instead of doing it with spaces either side of the text

MerrettM: Good point, although I think this was just how we wrote this rather than trying to add padding.

src/Button.test.js
```diff
@@ -0,0 +1,34 @@
...
+  });
+  it("renders a button with text", () => {
+    expect.assertions(1);
+
+    const result = shallow(<Button children="Test" />);
+
+    expect(result.debug()).toMatchSnapshot();
+  });
+  it("executes passed function when clicked", () => {
+    expect.assertions(1);
```
RaMorey: I think you only need assertions for asynchronous code

src/Button.test.js
```diff
@@ -0,0 +1,34 @@
...
+
+    const callBack = jest.fn();
+
+    const wrapper = shallow(<Button onClick={callBack} children="Test" />);
```
RaMorey: You want to pass the children inside of the button component `<Button>children</Button>

src/__snapshots__/Button.test.js.snap
```diff
@@ -0,0 +1,9 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Button renders a blank button 1`] = `"<styled.button onClick={[undefined]} />"`;
+
+exports[`Button renders a button with text 1`] = `
+"<styled.button onClick={[undefined]}>
```
RaMorey: I'd probably pass an onClick to each of the tests, as it would be a required prop

MerrettM: Would that mean we'd have to mock a function for each one? or just use some placeholder?

src/Button.test.js
MerrettM: So should we remove it from all the non-asynchronous tests?

src/__snapshots__/Button.test.js.snap
RaMorey: No, so you normally only use a jest.fn() mock when you are testing to see if that function was called. But when you're just providing a function and dont care if it's called or not you can just give it the value `() => {}`

src/Button.test.js
RaMorey: I think so but check with shannon as I don't use the expect.assertions stuff - So I might be wrong

shanassaf: I include them in every test purely for consistency's sake but entirely up to you

src/Button.test.js
```diff
@@ -0,0 +1,34 @@
...
+
+    expect(component.debug()).toMatchSnapshot();
+  });
+  it("executes passed function when clicked", () => {
+    expect.assertions(1);
+
+    const callBack = jest.fn();
+
+    const wrapper = shallow(<Button onClick={callBack}>Test</Button>);
+
```
shanassaf: I normally prefer to keep the 'act' lines together so no need for this space here. Entirely up to you, but either way don't bother removing in this PR.

src/Button.test.js
```diff
@@ -0,0 +1,34 @@
+import Button, { StyledButton } from "./Button";
+import { configure, shallow } from "enzyme";
+import Adapter from "enzyme-adapter-react-16";
+import React from "react";
+
```
shanassaf: Try importing 'jest-styled-components' and see how the snapshots look (it actually shows the css in the snapshot so if someone changes the styles the snapshot will tell you). 
https://github.com/styled-components/jest-styled-components
If you like it keep it, if you don't, remove it.

PR #8, opened 2019-10-17
Gp19 40 conference detail component

karenzamani:
This PR combines two tickets GP19-40 & GP19-41. A react router has been set-up and a test for conference details component has been created.


src/Conference/ConferenceDetails.jsx
```diff
@@ -0,0 +1,36 @@
...
+  width: 100%;
+  height: 400px;
+  color: white;
+  background: darkslategray;
+`;
+
+const ConferenceDetails = ({ conferences }) => {
+  let { id } = useParams();
+  let conference = {};
+  conferences.forEach(singleConference => {
```
dex-mendeley: Javascript arrays have a method called find that given a function returns the object(s) that pass this value.

So in your case you'd write:

const conference = conferences.find(conference => conference.id ===id)

src/App.js
```diff
@@ -34,11 +37,20 @@ function App() {
       {isLoading ? (
         <p>Loading...</p>
       ) : (
-        <ConferenceList conferences={conferences} />
+      <Router>
```
dex-mendeley: I'm not sure but I believe it might be better to move the Loading logic into the router so it would look something like that:

`<Router>
   {isLoading ? ( <p>Loading....</p>):(<Switch>......[Rest of switch markup]</Switch></Router>`

package.json
```diff
@@ -4,8 +4,10 @@
   "private": true,
   "dependencies": {
     "axios": "^0.19.0",
+    "history": "^4.10.1",
```
RaMorey: Pretty sure you can remove this package :)

src/App.js
```diff
@@ -1,6 +1,9 @@
 import React, { useState, useEffect } from "react";
 import styled from "styled-components";
-import ConferenceList from "./ConferenceList";
+import { BrowserRouter as Router, Switch, Route } from "react-router-dom";
+// import ConferencePage from "./Conference/ConferencePage";
```
RaMorey: We can remove this comment now

src/App.js
```diff
@@ -34,11 +37,20 @@ function App() {
...
       ) : (
-        <ConferenceList conferences={conferences} />
+      <Router>
+        <Switch>
+          <Route exact path="/">
+            <ConferenceList conferences={conferences} />
+          </Route>
+          <Route path="/:id">
+            <ConferenceDetails conferences={conferences} />
+          </Route>
```
RaMorey: If you guys like you can add a redirect at the end to account for any other url people might type in.https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Redirect.md

src/Conference/ConferenceDetails.jsx
RaMorey: Agreed, find would be nicer here :) Although you'll still need to convert the id to a number, using parseInt or Number as you already are.

src/index.js
```diff
@@ -1,10 +1,16 @@
...
+import React from "react";
+import ReactDOM from "react-dom";
+import "./index.css";
+import App from "./App";
+import * as serviceWorker from "./serviceWorker";
+import { BrowserRouter } from "react-router-dom";
 
-ReactDOM.render(<App />, document.getElementById('root'));
+ReactDOM.render(
+  <BrowserRouter>
```
RaMorey: I'm not sure it's good practice to have 2 lots of BrowserRouter in your application, it might cause some random bugs. I think it would be better to keep this one and remove the Router that you have inside of app.js? It would also solve Dex's query he raised

src/Conference/ConferenceDetails.jsx
karenzamani: Hi Dex, just a bit confused as to where the "const conference = conferences.find(conference => conference.id ===id)" should be used. We were assuming it was in place of the if statement?

src/App.js
karenzamani: Do you mean if someone types in the wrong URL they will be redirected back to the homepage?

RaMorey: Well currently they won't be, but if that is your desired outcome then yes you'd need a redirect.

package.json
```diff
@@ -4,8 +4,10 @@
   "private": true,
   "dependencies": {
     "axios": "^0.19.0",
+    "node": "^12.12.0",
```
RaMorey: Don't think you need this?

src/Conference/ConferenceItem.jsx
```diff
@@ -1,7 +1,12 @@
 import React from "react";
 import styled from "styled-components";
+<<<<<<< HEAD:src/Conference/ConferenceItem.jsx
```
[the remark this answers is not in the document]
RaMorey: Ya when you merge with master you have to resolve the conflicts as well guys, have a look in visual studio code and see what it's asking you to do

PR #9, opened 2019-10-21
Gp19 43 create nav bar

GBurley-Jones:


src/Navbar.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const StyledNavbar = styled.nav`
+  text-align: center;
+  background: papayawhip;
+`;
+
+const Navbar = () => {
+  return (
+    <StyledNavbar>
+      <div >
```
dex-mendeley: You can probably remove the div as it doesn't have any styling

src/Navbar.test.js
```diff
@@ -0,0 +1,31 @@
...
+});
+
+describe("Navbar", () => {
+    it("renders a navbar with text", () => {
+        expect.assertions(2);
+
+        const component = shallow(<Navbar />);
+
+        expect(component).toMatchSnapshot();
+        expect(component.find(NavLink).childAt(0).text()).toEqual('Home');
```
RaMorey: You don't really need this line as the snapshot kinda covers it

src/Navbar.jsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const StyledNavbar = styled.nav`
+  text-align: center;
+  background: papayawhip;
+`;
+
+const Navbar = () => {
+  return (
+    <StyledNavbar>
+      <NavLink to="/">Home</NavLink>
```
tonymaynard: Is there any reason you are using `NavLink` over `Link`? Theres nothing wrong with this but the docs say `NavLink` is used when you want to add styling for that given path in the nav bar ( https://reacttraining.com/react-router/web/api/NavLink) so perhaps you could add some styling for when this path is active?

src/Navbar.test.js
```diff
@@ -0,0 +1,31 @@
+import Navbar from "./Navbar";
+import { configure, shallow } from "enzyme";
+import Adapter from "enzyme-adapter-react-16";
+import { NavLink } from 'react-router-dom';
+import React from "react";
+
+configure({ adapter: new Adapter() });
+//adapter adapts enzyme for our version of react?
+
```
tonymaynard: Yeah so the adapter is used to create a bridge between the enzyme API which I don't think has really changed and the forever changing versions of react, with different versions of react they release different render methods and different ways to interact with the DOM which would interfere with the `shallow` and `mount` , so the adapters allow you to write tests normally depending on which version of react

sophiambolah: Perfect, that makes sense! Thank you

src/App.js
```diff
@@ -31,21 +33,24 @@ function App() {
...
+            <Switch>
+              <Route exact path="/">
+                <ConferenceList conferences={conferences} />
+              </Route>
+              <Route path="/:id">
+                <ConferenceDetails conferences={conferences} />
+              </Route>
+            </Switch>
+          )}
+        {errorCaught ? <p>An error has occurred...</p> : ""}
```
dex-mendeley: Another way (maybe a little bit more elegant) to write this is { errorCaught && <p>An error has occurred...</p> } so as long as there's no error nothing will be rendered and as soon is there's one it will render the <p>.

Doesn't have to be changed as part of the PR or at all though ;)

sophiambolah: ok thank you!! :)

PR #10, opened 2019-10-22
Gp19 45 get conference endpoint

GBurley-Jones:
We had a few git issues! Sorry for all of the commits. The final commit works and contains everything we want it to. 
Thanks!


src/api.test.js
```diff
@@ -40,3 +40,29 @@ describe("getConferenceList", () => {
...
+        "From Festivalofmarketing.com: The Festival of Marketing is a unique experience where ambitious marketers can discover, learn, celebrate and shape the future together. As the largest global event dedicated to brand marketers, the Festival reflects the very nature of ...",
+      topic: "Marketing"
+    }
+  };
+
+  //test to return conference by id
+  it("returns conference by id", async () => {
+    expect.assertions(1);
+
+    axios.get.mockResolvedValue(mockData);
```
RaMorey: This test is only really testing whether the mockResolvedValue works, which we can assume it does. I'm not totally sure we need a test for this file

src/Conference/ConferenceDetails.jsx
```diff
@@ -10,9 +11,41 @@ const StyledConferenceDetails = styled.div`
...
+const ConferenceDetails = () => {
+
+  const [isLoading, setIsLoading] = useState(true);
+  const [conference, setConference] = useState(null);
+  const [errorCaught, setErrorCaught] = useState(false);
+  const { id } = useParams();
+
+
+  useEffect(() => {
+    const fetchData = async () => {
```
tonymaynard: For readability you could take the `fetchData` function out of the useEffect

src/Conference/__snapshots__/ConferenceDetails.test.js.snap
```diff
@@ -1,27 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`ConferenceDetails renders conference details 1`] = `
```
tonymaynard: I don't think you should've updated this snapshot as its now not showing the conference details! I think you should perhaps try to mock the request and resolve the promise so it sets some data in the state and renders the component as opposed to just the loading state

sophiambolah: We are a bit unsure on how to proceed with this, could you offer any further explanation? Thanks!

src/Conference/ConferenceDetails.test.js
```diff
@@ -1,8 +1,14 @@
 import ConferenceDetails from "./ConferenceDetails";
-import { configure, shallow } from "enzyme";
+import { configure, shallow, mount } from "enzyme";
```
dex-mendeley: Forgot to mention earlier, you're not using shallow anymore, so you can remove the import here ;)

src/Conference/ConferenceDetails.test.js
```diff
@@ -13,28 +19,45 @@ jest.mock('react-router-dom', () => {
...
-    useParams.mockReturnValue({ id: '1'})
- 
-    const component = shallow(<ConferenceDetails conferences={mockData} />);
-    
-    expect(component.debug()).toMatchSnapshot();
+    axios.get.mockResolvedValue(mockData);
+
+    useParams.mockReturnValue({ id: '1' });
+    let component;
+    component = mount(<ConferenceDetails />);
```
dex-mendeley: Just a little post-refactor cleanup: you don't need to declare and assign in two lines anymore, you can now just assign to a const straight away :)

src/api.js
```diff
@@ -9,3 +9,13 @@ export const getConferenceList = async () => {
     throw error;
   }
 };
+
+export const getConferenceById = async (id) => {
+  try {
+    const response = await axios.get(`http://localhost:8080/conferences/${id}`);
```
dex-mendeley: Maybe something for future PRs: All these URLs probably should be put in to a config file where they can be changed and maintained more easily :) but not really part of this PR :)

src/Conference/ConferenceDetails.test.js
```diff
@@ -17,36 +19,45 @@ jest.mock('react-router-dom', () => {
 
 configure({ adapter: new Adapter() });
 
+beforeAll(() => {
```
dex-mendeley: Any reason why you removed the comment? Another developer might not understand why we're suppressing
warning at this point now :)

src/api.js
GBurley-Jones: ok we'll look into doing this in the future, thank you!!

src/Conference/ConferenceDetails.test.js
GBurley-Jones: Sorry the comment is still there in ours! We probably just didn't save it before we pushed, we'll push again!

PR #11, opened 2019-10-22
Grads 0 styling changes

GBurley-Jones:


src/Conference/ConferenceDetails.jsx
```diff
@@ -6,10 +6,39 @@ import { Link, useParams } from "react-router-dom";
...
+vertical-align: middle;
+
+
 `;
 
+const StyledDetailsCard = styled.div`
+  border-style: solid;
+  border-color: black;
+  border-radius: 25px;
+  background: ${props => props.background || "#dbd8db"};
```
tonymaynard: I'm not sure you need props here as you're not passing any props to the component!

src/App.js
```diff
@@ -7,9 +7,12 @@ import { getConferenceList } from "./api";
 import Navbar from "./Navbar";
 import { BrowserRouter } from "react-router-dom";
 
-const StyledApp = styled.div`
+const StyledApp = styled.body`
   text-align: center;
-  background: lightgrey;
+  background: #ccb7bb;
+  width: auto;
+  height: 1000px;
```
tonymaynard: out of interest, why have you set the height to 1000px for the body?

src/Conference/ConferenceItem.jsx
```diff
@@ -3,38 +3,53 @@ import styled from "styled-components";
...
+  width: 315px;
+  height: 250px;
   transition: 0.3s;
-  background: ${props => props.background || "papayawhip"};
-
-  :hover {
-    box-shadow: 0 8px 16px 0 rgba(0, 0, 0, 0.2);
-  }
+  :hover {box-shadow: 0 8px 16px 0 rgba(0, 0, 0, 0.2);}
+  vertical-align: middle;
```
tonymaynard: Could use flex here instead to position the cards

src/Navbar.jsx
```diff
@@ -3,14 +3,19 @@ import styled from "styled-components";
 import { Link } from 'react-router-dom';
 
 export const StyledNavbar = styled.nav`
-  text-align: center;
-  background: papayawhip;
-`;
+  background: #ccb7bb;
```
tonymaynard: Probably shouldn't be the same colour as the background

src/Navbar.jsx
```diff
@@ -3,14 +3,19 @@ import styled from "styled-components";
...
+  width: 100%;
+  height: 30px;
+  `;
+
+export const StyledLink = styled(Link)`
```
tonymaynard: May want to position this witihin the navbar, usually home buttons are positioned to the left

src/App.js
sophiambolah: We have set the height of the background to make sure it covers the whole page and not just the objects within it, is this okay?

tonymaynard: Have a look at `vh` I prefer using this as it allows you to set the height to the users screen

src/Navbar.jsx
```diff
@@ -3,14 +3,21 @@ import styled from "styled-components";
 import { Link } from 'react-router-dom';
 
 export const StyledNavbar = styled.nav`
-  text-align: center;
-  background: papayawhip;
-`;
+  background: #322d38;
+  width: 100%;
```
dex-mendeley: If you're working with 100vh for the body height, it make sense to use 100vw here for the width :)

src/Conference/ConferenceItem.jsx
```diff
@@ -3,38 +3,51 @@ import styled from "styled-components";
...
+  width: 315px;
+  height: 250px;
   transition: 0.3s;
-  background: ${props => props.background || "papayawhip"};
-
-  :hover {
-    box-shadow: 0 8px 16px 0 rgba(0, 0, 0, 0.2);
-  }
+  :hover {box-shadow: 0 8px 16px 0 rgba(0, 0, 0, 0.2);}
+  .container {
```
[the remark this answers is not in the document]
GBurley-Jones: We have tried without the .container but it seems to mess up all of the formatting of the text inside the cards, we're not sure how to implement it without this?

PR #14, opened 2019-10-31
Gp19 51 create form

MerrettM:


src/Conference/AddConference.jsx
```diff
@@ -0,0 +1,46 @@
...
+  }
+  vertical-align: middle;
+  margin: 0 auto;
+`;
+
+const AddConference = () => {
+  return (
+    <StyledConferenceForm>
+      <StyledCard>
+        <StyledCardHeading>ADD A NEW CONFERENCE:</StyledCardHeading>
```
RaMorey: Personally I'd capitalize this text using css

src/Conference/ConferenceList.test.js
```diff
@@ -1,11 +1,36 @@
...
+import { act } from "react-dom/test-utils";
+import { createMemoryHistory } from "history";
 
 configure({ adapter: new Adapter() });
 
+jest.mock("../api", () => ({
+  getConferenceList: jest.fn()
+}));
+
+/* This is a temporary workaround for enzyme/react bug #14769 (github.com/facebook/react/issues/14769):
```
j-hutchinson: if this is all commented out, remove it

src/Form.jsx
```diff
@@ -0,0 +1,95 @@
+import React, { useState } from "react";
+import Input from "./Input";
+import { createNewConference } from "./api";
+import { useHistory } from "react-router-dom";
+
+const Form = () => {
+  let history = useHistory();
+  const [state, setState] = useState({
```
j-hutchinson: `state` should `conference`, this is a variable on state rather than the entire state object. Naming convention here could get confusing.
Then `setConference`

src/Form.jsx
```diff
@@ -0,0 +1,95 @@
...
+  };
+
+  const submitForm = async () => {
+    const newState = {
+      ...state,
+      dateTime: state.dateTime + ":00Z"
+    };
+    try {
+      await createNewConference(newState);
+      history.push("/"); // navigate to home page 😃
```
j-hutchinson: remove comment

src/Form.jsx
```diff
@@ -0,0 +1,95 @@
...
+        ev.preventDefault();
+        submitForm();
+      }}
+    >
+      <Input
+        id="name"
+        label="Conference Name: "
+        type="text"
+        name="name"
+        value={state.name}
```
j-hutchinson: then `conference.name` and throughout

package.json
```diff
@@ -32,6 +32,8 @@
     ]
   },
   "devDependencies": {
+    "@types/history": "^4.7.3",
+    "@types/jest": "^24.0.21",
```
tonymaynard: Are these dependencies needed?

src/Conference/ConferenceDetails.test.js
```diff
@@ -48,20 +52,19 @@ describe("ConferenceDetails", () => {
...
     axios.get.mockResolvedValue(mockData);
 
-    useParams.mockReturnValue({ id: '1' });
+    useParams.mockReturnValue({ id: "1" });
 
     const component = mount(<ConferenceDetails />);
 
     setImmediate(() => {
       component.update();
       expect(component.debug()).toMatchSnapshot();
```
tonymaynard: fyi I don't think you need to use `.debug()` anymore as you have a serialiser in your package.json :)

src/Form.jsx
```diff
@@ -0,0 +1,95 @@
...
+        name="name"
+        value={conference.name}
+        onChange={handleChange}
+        required
+        maxLength="80"
+      />
+      <br />
+      <Input
+        label="Date and Time: "
+        type="datetime-local"
```
tonymaynard: Have you tried using this on safari or firefox? I'm not 100% sure it is supported may be worth checking https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local

src/Form.jsx
```diff
@@ -0,0 +1,95 @@
...
+
+  return (
+    <form
+      onSubmit={ev => {
+        ev.preventDefault();
+        submitForm();
+      }}
+    >
+      <Input
+        id="name"
```
tonymaynard: May want to remove this id and the one on the submit we tried for testing purposes!

src/Form.jsx
MerrettM: Just did it, not supported by safari, I don't have Firefox to check that though

package.json
MerrettM: I think those were added by Sethi to help us work out what we needed to do for testing

src/Form.jsx
MerrettM: Harry has said we don't have to support any browsers other than Chrome so I guess this is okay

package.json
tonymaynard: okay just never seen them before! thanks :)

src/Form.jsx
tonymaynard: Yeah this isn't a problem for this project, just wanted to highlight how different browsers are, its not a blocker

PR #15, opened 2019-11-01
GP19-53 create delete button

karenzamani:


src/Conference/ConferenceDetails.jsx
```diff
@@ -1,8 +1,9 @@
 import React, { useEffect, useState } from "react";
 import styled from "styled-components";
 import { getTimestring, getDatestring } from "../utils";
-import { Link, useParams } from "react-router-dom";
-import { getConferenceById } from "../api.js";
+import { Link, useParams, useHistory } from "react-router-dom";
```
dex-mendeley: Bit of a nit: Usually you'd order your imports: 

1. External libraries
2. Your components
So I'd put react-router-dom before the utils import :)

src/Conference/ConferenceDetails.jsx
```diff
@@ -74,6 +76,15 @@ const ConferenceDetails = () => {
   const { name, topic, dateTime, city, description } = conference || {};
   const date = new Date(dateTime);
 
+  const deleteAConference = async id => {
+    try {
+      await deleteConference(id);
+      history.push("/");
+    } catch (error) {
+      console.log(error);
```
dex-mendeley: Just out of interest: Are you planning on doing something in the future if there's an error like like a notification?

src/Conference/ConferenceDetails.jsx
```diff
@@ -84,6 +95,14 @@ const ConferenceDetails = () => {
         <p>{city}</p>
         <p>{description}</p>
         <StyledLink to="/">Back</StyledLink>
+        <br />
+        <Button
+          onClick={() => {
+            deleteAConference(id);
```
dex-mendeley: A small performance improvement: I'd hoist this out of the render method as every time this re-renders you create a new reference. :)

src/Form.jsx
```diff
@@ -0,0 +1,82 @@
...
+  const [state, setState] = useState({
+    name: "",
+    dateTime: date, // fix format of this to match BE
+    city: "",
+    description: "",
+    topic: ""
+  });
+
+  const handleChange = event => {
+    const value = event.target.value;
```
dex-mendeley: You could destructure the event, so you wouldn't have to type event.target.x twice ;)

src/Form.jsx
```diff
@@ -0,0 +1,82 @@
...
+    setState({ ...state, [event.target.name]: value });
+  };
+
+  const submitForm = async () => {
+    console.log(state);
+    const newState = {
+      ...state,
+      dateTime: "2020-06-02T21:34:33.616Z"
```
tonymaynard: I don't think this should be hard coded!

src/Form.jsx
```diff
@@ -0,0 +1,82 @@
+import React, { useState } from "react";
```
tonymaynard: Should this file be in this PR I thought it's already merged?

src/api.js
```diff
@@ -20,9 +20,23 @@ export const getConferenceById = async (id) => {
   }
 };
 
-export const createNewConference = async (conference) => {
+export const createNewConference = async conference => {
   try {
-    const response = await axios.post(`http://localhost:8080/conferences, ${conference}`);
+    const response = await axios.post(
+      `http://localhost:8080/conferences, ${conference}`
```
tonymaynard: should probably be written as `post('http://localhost:8080/conferences', conference)`

MerrettM:
This branch was created off the incomplete create a form branch, we are currently working on resolving the conflicts and bringing it up to date with the master, I've resolved all the conversations relating to these issues as we've fixed them in the commit we're currently working on.

src/Conference/ConferenceDetails.jsx
MerrettM: Next ticket is a toast component which might be used for this? currently we're just console logging here as we haven't really figured out how we want to do error handling.

src/Conference/ConferenceDetails.jsx
```diff
@@ -73,6 +75,18 @@ const ConferenceDetails = () => {
...
 
+  const deleteAConference = async id => {
+    try {
+      await deleteConference(id);
+      history.push("/");
+    } catch (error) {
+      console.log(error);
+    }
+  };
+  const onClick = () => {
```
tonymaynard: I probably would be more declerative with what its actually supposed to do, perhaps `deleteConference` so you can see when glancing at the button what it actually does

src/Conference/ConferenceDetails.test.js
```diff
@@ -1,28 +1,34 @@
...
+import { createMemoryHistory } from "history";
+import { act } from "react-dom/test-utils";
 
 jest.mock("react-router-dom", () => {
+  const originalReactRouter = jest.requireActual("react-router-dom");
   return {
-    useParams: jest.fn(),
+    ...originalReactRouter,
+    useParams: jest.fn(() => ({ id: "1" })),
+    //useHistory: jest.fn(),
```
tonymaynard: If this is no longer needed I would remove

src/Conference/ConferenceDetails.test.js
```diff
@@ -41,30 +46,48 @@ afterEach(() => {
...
+      );
 
-    setImmediate(() => {
-      component.update();
-      expect(component).toMatchSnapshot();
-      done();
+      await apiReturnValue;
     });
+    wrapper.update();
+    expect(wrapper.containsMatchingElement(<h3>{mockData.name}</h3>)).toEqual(
```
j-hutchinson: I think a better way of testing this is to add classnames to the html elements. Then you can find those in wrapper ie `expect(wrapper.find(className).length).toBe(1)` you can also check the text inside the element to match what you expect

src/Conference/ConferenceDetails.jsx
```diff
@@ -83,6 +97,8 @@ const ConferenceDetails = () => {
         <p>{city}</p>
```
j-hutchinson: so these would become `<p className="city" />` for example

src/api.js
```diff
@@ -23,7 +23,7 @@ export const getConferenceById = async id => {
 export const createNewConference = async conference => {
   try {
     const response = await axios.post(
-      "http://localhost:8080/conferences",
+      `http://localhost:8080/conferences`,
```
RaMorey: Is there a reason this string is not using normal quotes?

src/Conference/ConferenceDetails.test.js
```diff
@@ -41,30 +46,40 @@ afterEach(() => {
...
-      done();
+      await apiReturnValue;
     });
+    wrapper.update();
+    expect(wrapper.find("h3.name").length).toBe(1);
+    expect(wrapper.find(".topic").length).toBe(1);
+    expect(wrapper.find(".date").length).toBe(1);
+    expect(wrapper.find(".time").length).toBe(1);
+    expect(wrapper.find(".city").length).toBe(1);
+    expect(wrapper.find(".description").length).toBe(1);
```
tonymaynard: I Think this is a good change, but I still think as you are returning the mocked data you should still check that the correct data is being displayed within the elements, currently all you have checked is that the element exists not the content. If someone told you just to test this please feel free to ignore

src/api.js
karenzamani: Hi Ross, do you mean the post request should be in back ticks? (thats how it looks in our version) or are you suggesting we change to ""??

RaMorey: Normally you only use back ticks for when you are concatenating variables and strings together. So if it's only a string I'd use normal ""

j-hutchinson: for context thats using template literal strings 👍

src/Conference/ConferenceDetails.test.js
j-hutchinson: We have made the decision to start lightly testing components such as these, due to time constraints. If we test everything as much as we would expect the FE will fall further behind 😔

tonymaynard: okay makes sense 👍

src/Conference/ConferenceDetails.test.js
```diff
@@ -41,30 +46,40 @@ afterEach(() => {
...
+      );
 
-    setImmediate(() => {
-      component.update();
-      expect(component).toMatchSnapshot();
-      done();
+      await apiReturnValue;
     });
+    wrapper.update();
+    expect(wrapper.find("h3.name").length).toBe(1);
```
j-hutchinson: You dont need `h3` in this find. Its a bit verbose but it will still pass

karenzamani: Hi Jack, when we remove the h3 the test fails saying received 2 instead of 1. We aren't sure why this is happening

src/App.js
```diff
@@ -31,6 +33,7 @@ function App() {
           </Route>
         </Switch>
       </StyledApp>
+      <ToastContainer position="bottom-right" autoClose={5000} pauseOnHover/>
```
RaMorey: I think autoClose=5000 and pauseOnHover are the defaults, you can leave them if you think it's more readable though

src/Conference/ConferenceDetails.jsx
```diff
@@ -73,16 +75,30 @@ const ConferenceDetails = () => {
   const { name, topic, dateTime, city, description } = conference || {};
   const date = new Date(dateTime);
 
+  const deleteConference = async id => {
+    try {
+      await deleteConferenceById(id);
+      history.push("/");
```
RaMorey: You want to put the toast emitter in here, and maybe another one in the catch for the error :)

RaMorey: Actually just realised you did it in the api file, up to you :) But you didn't need the pauseOnHover bits in that file btw

src/Conference/ConferenceDetails.test.js
j-hutchinson: Interesting, it's finding another element (not a h3) with classname 'name'. Do you have this somewhere else?

src/api.js
```diff
@@ -23,11 +24,23 @@ export const getConferenceById = async id => {
...
+    throw error;
+  }
+};
+
+export const deleteConferenceById = async id => {
+  try {
+    const response = await axios.delete(
+      `http://localhost:8080/conferences/${id}`
     );
+    toast.success("Conference successfully deleted!", { pauseOnHover: true });
```
tonymaynard: I don't think the `pauseOnHover` is needed as it is inherited from the container

tonymaynard:
In future try not to mix tickets into one PR if possible, it makes the reviews longer and larger if people haven't necessarily seen the previous ticket.

src/Conference/ConferenceDetails.test.js
karenzamani: The only place we reference 'name' is in conference, we don't use 'className = name' anywhere else as far as we can see

longstaff-elsevier: should be introducing data-test tags for specificity?

j-hutchinson: @longstaff-elsevier we could do that, i feel like a specific className is enough at this point or alternatively if the hmtl element is unique you can import the styled component and find it

PR #17, opened 2019-11-06
Gp19-64: update conference page

MerrettM:
Also includes:
GP19-59: edit conference toast
GP19-57: edit conference button


RaMorey:
I personally think it would be much nicer to pull the `getConference` logic up a component and share the response with both conferenceDetails and conferenceUpdate. At the moment you have to go to conferenceDetails (make the GET call), then click edit (make another get call for the same information you just did). 

I think it would be nicer to make the call once and pass it to both

src/Conference/UpdateConference.jsx
```diff
@@ -0,0 +1,153 @@
...
+    }
+  };
+
+  useEffect(() => {
+    if (id) {
+      fetchData(id);
+    }
+  }, [id]);
+
+  const handleChange = event => {
```
RaMorey: Similar to the other pr, you can deconstruct event to name and value to save rewriting it each time

src/Conference/UpdateConference.jsx
```diff
@@ -0,0 +1,153 @@
...
+    const value = event.target.value;
+    setConference({ ...conference, [event.target.name]: value });
+  };
+
+  const submitForm = async () => {
+    const newConference = {
+      ...conference,
+      dateTime: conference.dateTime + "Z"
```
RaMorey: This seems odd, do you really need to do it? I've not got time to look at it properly right now but if you really need to do it then I'd use a template literal.

MerrettM: Unfortunately due to how date time-local outputs a date we do, or it'll be refused by the back end as it doesn't fit the Instant format

MerrettM: but I agree with the template literal

MerrettM:
> I think it would be nicer to make the call once and pass it to both

Definitely agree with you @RaMorey we did mean to change that after we got the form working so we'll do that in the next commit.

src/Conference/UpdateConference.jsx
```diff
@@ -0,0 +1,153 @@
...
+
+const StyledCard = styled.div`
+  border-style: solid;
+  border-color: black;
+  border-radius: 25px;
+  background: #dbd8db;
+  width: 65%;
+  height: 370px;
+  transition: 0.3s;
+  :hover {
```
longstaff-elsevier: prefer these to be seperated underneath generic style rather than halfway through, makes readability better

src/Conference/UpdateConference.jsx
```diff
@@ -0,0 +1,153 @@
...
+    box-shadow: 0 8px 16px 0 rgba(0, 0, 0, 0.2);
+  }
+  vertical-align: middle;
+  margin: 0 auto;
+`;
+
+const UpdateForm = () => {
+  let history = useHistory();
```
longstaff-elsevier: why let here?

src/Conference/UpdateConference.jsx
```diff
@@ -0,0 +1,153 @@
...
+          <Input
+            label="Conference Name: "
+            type="text"
+            name="name"
+            value={conference.name}
+            onChange={handleChange}
+            required
+            maxLength="80"
+          />
+          <br />
```
longstaff-elsevier: Generally probably shouldn't use <br>, either wrap Input in block element like <p> or use CSS to break the line

src/Conference/UpdateConference.jsx
MerrettM: We do a history.push later and assumed that for this to work history would need to be a let so it could be changed, is this not the case?

src/Conference/ConferenceCall.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const fetchData = async conferenceId => {
+    setIsLoading(true);
+
+    try {
+      const conference = await getConferenceById(conferenceId);
+      setConference(conference);
+      setUpdatedConference({
+        name: conference.name,
+        dateTime: conference.dateTime.substring(
```
tonymaynard: why do you need to manipulate the `dateTime` you get back from the API?

MerrettM: Because the backend uses `YYYY-MM-DDTHH:MM:SSZ` and date time local seems to require `YYYY-MM-DDTHH:MM:SS` to work

src/Conference/UpdateConference.jsx
tonymaynard: As we want to keep the same history object throughout I don't think this needs to be `let` as we are not reassigning the variable `history`, history.push is just using a method on the history object.

src/Conference/ConferenceCall.jsx
dex-mendeley: This works, in case this gets unwieldy in the future you could also introduce a library such as momentjs (https://momentjs.com/) to handle this but probably not needed here :)

src/Conference/ConferenceCall.jsx
```diff
@@ -0,0 +1,64 @@
+import React, { useState, useEffect } from "react";
+import { Switch, Route, useParams } from "react-router-dom";
+import ConferenceDetails from "./ConferenceDetails";
+import UpdateConference from "./UpdateConference";
+import { getConferenceById } from "../api.js";
+
+const ConferenceCall = () => {
```
dex-mendeley: I'm not quite sure about the name ConferenceCall. In React parlance you'd probably call this a ConferenceContainer.

PR #18, opened 2019-11-11
GP19-12: create registration page, link in navbar and axios request

GBurley-Jones:


src/User/RegistrationForm.jsx
```diff
@@ -0,0 +1,138 @@
...
+
+const StyledCard = styled.div`
+  border-style: solid;
+  border-color: black;
+  border-radius: 25px;
+  background: #dbd8db;
+  width: 65%;
+  height: 370px;
+  transition: 0.3s;
+  :hover {
```
j-hutchinson: To improve readability, move element states to the bottom of this css. Keeps all the base css together

src/api.js
```diff
@@ -24,7 +24,9 @@ export const getConferenceById = async id => {
 export const createNewConference = async conference => {
   try {
     const response = await axios.post(
-      "http://localhost:8080/conferences",conference);
+      "http://localhost:8080/conferences",
+      conference
```
j-hutchinson: should conference not be wrapped in {} as body is an object?

src/App.js
```diff
@@ -28,12 +29,15 @@ function App() {
           <Route path="/add">
             <AddConference />
           </Route>
+          <Route path="/users">
+            <RegistrationForm />
```
tonymaynard: Does `/users` signify a registration? Perhaps `/user/regiser`, initally `/users` suggest to me you want to list other users

src/User/RegistrationForm.jsx
```diff
@@ -0,0 +1,138 @@
...
+  const submitForm = async () => {
+    const newUser = {
+      ...user
+    };
+
+    try {
+      await createNewUser(newUser);
+      history.push("/");
+    } catch (error) {
+      console.log(error);
```
tonymaynard: Maybe a better use would be a toast to show the user

src/User/RegistrationForm.jsx
```diff
@@ -0,0 +1,138 @@
...
+            required
+            maxLength="100"
+          />
+          <Input
+            label="Password: "
+            type="text"
+            name="password"
+            value={user.password}
+            onChange={handleChange}
+            required
```
tonymaynard: Might be worth using a regex pattern to help force the user to create a more secure password :) instead of just `password: a` which would pass

src/User/RegistrationForm.jsx
```diff
@@ -0,0 +1,138 @@
...
+      console.log(error);
+    }
+  };
+
+  return (
+    <StyledUserForm>
+      <StyledCard>
+        <StyledCardHeading>add a new user:</StyledCardHeading>
+        <form
+          onSubmit={ev => {
```
dex-mendeley: Small performance improvement: maybe hoist the submitHandler similar to how you're already doing it with the change handler

src/User/RegistrationForm.jsx
```diff
@@ -0,0 +1,138 @@
...
+            type="text"
+            name="username"
+            value={user.username}
+            onChange={handleChange}
+            required
+            maxLength="100"
+          />
+          <Input
+            label="Password: "
+            type="text"
```
dex-mendeley: As this is a password input, it probably makes sense to set the input type to 'password'

src/api.js
GBurley-Jones: We have tried with the bracket added in, but it stops the API request working?

src/User/RegistrationForm.jsx
GBurley-Jones: This wasn't in the ticket but we are happy to add it in now!

src/User/RegistrationForm.jsx
GBurley-Jones: This is a separate ticket, we will implement this on another branch :)

src/api.js
j-hutchinson: Conference is an object so it should be fine actually, ignore me 😅

src/Conference/UpdateConference.jsx
```diff
@@ -0,0 +1,124 @@
...
+import styled from "styled-components";
+import { useHistory } from "react-router-dom";
+import { updateConferenceById } from "../api";
+
+const StyledConferenceForm = styled.div`
+  height: 400px;
+  padding: 20px;
+`;
+
+const StyledCardHeading = styled.h3`
```
tonymaynard: As you are using this headings and styled cards throughout different components it may make sense to create a new file and import these into everywhere which needs them

src/StyledFormComponents.js
```diff
@@ -0,0 +1,34 @@
+import styled from "styled-components";
+
+const StyledForm = styled.div`
+  height: 400px;
+  padding: 20px;
+`;
+
+const StyledCardHeading = styled.h3`
```
tonymaynard: can you not now import these two?

src/StyledFormComponents.js
```diff
@@ -0,0 +1,34 @@
...
+  background: #dbd8db;
+  width: 100%;
+  height: 55px;
+  font-size: 30px;
+  font-weight: bold;
+  text-align: center;
+  vertical-align: middle;
+  text-transform: uppercase;
```
tonymaynard: nit: Should headers be all uppercase? Maybe capitalize is a less aggressive approach!

shanassaf:
Could you add a screenshot please? Just so we can see your beautiful page ;)

sophiambolah:
Here you go!

Screenshot 2019-11-11 at 16 59 17

PR #19, opened 2019-11-11
GP19-74: username and password validation

sophiambolah:


KaramKapoor:
> My regex knowledge is a little rusty but wouldn't that pattern prevent emails like [abc.def@gmail.com](mailto:abc.def@gmail.com)?
Fixed it! Now it takes in test.test@gmail.com and test.test@gmail.co.uk / gmail.edu.org.uk and similar domains.

src/User/RegistrationForm.jsx
```diff
@@ -85,6 +87,8 @@ const RegistrationForm = () => {
             label="Username: "
             type="text"
             name="username"
+            pattern="^[a-zA-Z0-9]*$"
+            title="Must not contain any spaces."
```
dex-mendeley: This also doesn't allow special characters, so you might want to put that in the error message as well :)

src/User/RegistrationForm.jsx
```diff
@@ -65,8 +65,10 @@ const RegistrationForm = () => {
           />
           <Input
             label="Email Address: "
-            type="text"
+            type="email"
             name="email"
+            pattern="^[a-zA-Z0-9\.]+@[a-zA-Z0-9]+\.[\.A-Za-z]{1,10}"
```
dex-mendeley: I don't think you'll need the anchor elements (^ and $) here but if you use them make sure you use both of them (there's not $ in this pattern)

denishowew: You don't like the [default pattern](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email) for type=email?  Yours is quite restrictive.

KaramKapoor: We tried it without the custom regex pattern, but the default setting for type=email doesn't require a ".com" domain after the "@email".

PR #20, opened 2019-11-18
Gp19 14 login page

MerrettM:
Adds login form and creates sessionToken cookie.

### DIFF
--- a/src/Conference/ConferenceDetails.jsx
+++ b/src/Conference/ConferenceDetails.jsx
@@ -14,6 +14,12 @@ export const StyledLink = styled(Link)`
   color: #7a517d;
 `;
 
+export const StyledDescription = styled.p`
+  padding-left: 100px;
+  padding-right: 100px;
+  padding-bottom: 50px;
+`;
+
 const ConferenceDetails = ({ conference, id, isLoading, errorCaught }) => {
   const history = useHistory();
 
@@ -44,13 +50,14 @@ const ConferenceDetails = ({ conference, id, isLoading, errorCaught }) => {
     <StyledForm>
       <StyledCard>
         <StyledCardHeading className="name">{name}</StyledCardHeading>
+        <StyledLink to={`/${id}/edit`}>Edit Conference</StyledLink>
         <p className="topic">{topic}</p>
         <p className="date">{getDatestring(date)}</p>
         <p className="time">{getTimestring(date)}</p>
         <p className="city">{city}</p>
-        <p className="description">{description}</p>
-        <StyledLink to={`/${id}/edit`}>Edit Conference</StyledLink>
-        <br />
+        <StyledDescription className="description">
+          {description}
+        </StyledDescription>
         <Button onClick={deleteThisConference}>Delete Conference</Button>
       </StyledCard>
     </StyledForm>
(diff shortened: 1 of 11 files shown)


src/User/LoginForm.jsx
```diff
@@ -0,0 +1,77 @@
...
+    password: ""
+  });
+
+  const handleChange = event => {
+    const { name, value } = event.target;
+    setUser({ ...user, [name]: value });
+  };
+
+  const submitForm = async () => {
+    const newUser = {
```
RaMorey: whats the purpose of reassigning this?

MerrettM: There doesn't appear to be one so we've just removed that line.

longstaff-elsevier:
As commented in last retro, please be more explicit in naming

src/Conference/ConferenceDetails.jsx
```diff
@@ -14,6 +14,12 @@ export const StyledLink = styled(Link)`
   color: #7a517d;
 `;
 
+export const StyledDescription = styled.p`
+  padding-left: 100px;
+  padding-right: 100px;
+  padding-bottom: 50px;
```
tonymaynard: FYI This is the same as `padding: 0 100px 50px 100px`

src/Conference/ConferenceDetails.test.js
```diff
@@ -80,6 +80,7 @@ describe("ConferenceDetails", () => {
     expect(wrapper.find(".date").length).toBe(1);
     expect(wrapper.find(".time").length).toBe(1);
     expect(wrapper.find(".city").length).toBe(1);
-    expect(wrapper.find(".description").length).toBe(1);
+    expect(wrapper.find(".description").length).toBe(2);
+    //Description is length 2 because of the Styled Component used on it
```
longstaff-elsevier: This is a little bit of a flag, should this not be fixed rather than leaving a comment?

src/User/LoginForm.jsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const handleChange = event => {
+    const { name, value } = event.target;
+    setUser({ ...user, [name]: value });
+  };
+
+  const submitForm = async () => {
+    try {
+      const token = await loginUser(user);
+      setCookie("sessionToken", token, { path: "/" });
```
longstaff-elsevier: I would abstract this out a level, so that this form does not need to know what the cookie information is, only the token.

eg pass from parent:
`setToken = {token => setCookie("sessionToken", token, { path: "/" });}`

here:
`setToken(token)`

longstaff-elsevier: This means that the component is less tied to the implementation (cookies) and the constants such as the name of the token can be kept in the same file, improving maintainability

sophiambolah: ok, done! :)

src/Conference/ConferenceDetails.test.js
MerrettM: > This is a little bit of a flag, should this not be fixed rather than leaving a comment?

This element is the only one in the set of p tags to use styled components and I think that's what's causing it to register as length two since it is technically nested in the styled components tag. Is there a way to specifically assign the description tag to the p tags only and not the styled part?

MerrettM: > Also no snapshot for the conference details component?

I think we stopped using them because Sethi recommended it as they break every time you change something on the page even if it still looks the same.

longstaff-elsevier:
Looks good, might be one you want to revisit when you get to the new testing lib just to see how it differs in targeting

PR #21, opened 2019-11-18
GP19-89 fix update conference bug

sophiambolah:


src/Conference/ConferenceContainer.jsx
```diff
@@ -47,6 +47,8 @@ const ConferenceContainer = () => {
           id={id}
           setConference={setConference}
           setUpdatedConference={setUpdatedConference}
+          isLoading={isLoading}
+          errorCaught={errorCaught}
```
dex-mendeley: Nit: It's probably fine just to name this error, errorCaught seems unnecessary ;)

PR #22, opened 2019-11-19
GP19-83/84 create logout button which deletes cookie and sends axios request

sophiambolah:


src/App.js
```diff
@@ -24,12 +24,15 @@ function App() {
   const setToken = token => {
     setCookie("sessionToken", token, { path: "/" });
   };
+  const deleteToken = () => {
+    removeCookie("sessionToken", { path: "/" });
```
longstaff-elsevier: you might want to move "sessionToken" and "path" into a constant instead of duplicating here?

src/Navbar.jsx
```diff
@@ -24,7 +26,34 @@ export const StyledLink = styled(Link)`
...
+    try {
+      await logoutUser(token);
+      deleteToken();
+      history.push("/");
+    } catch (error) {
+      console.log(error);
+    }
+  };
+
+  const changeLoginStatus = () => {
```
longstaff-elsevier: would probably prefer to see this inline or as a separate component, this seems halfway between.

src/Navbar.jsx
```diff
@@ -37,9 +66,7 @@ const Navbar = () => {
         <StyledListItem>
           <StyledLink to="/users/register">Register</StyledLink>
         </StyledListItem>
-        <StyledListItem>
-          <StyledLink to="/users/login">Login</StyledLink>
-        </StyledListItem>
+        <StyledListItem>{changeLoginStatus()}</StyledListItem>
```
longstaff-elsevier: inline?
`{token ? <Button onClick={() => logout()}>Logout</Button> : <StyledLink to="/users/login">Login</StyledLink>}`

src/App.js
MerrettM: like this:
`const name = "sessionToken";
  const path = "/";
  const setToken = token => {
    setCookie(name, token, { path: path });
  };`
or this:
` const sessionToken = { name: "sessionToken", path: "/" };
  const setToken = token => {
    setCookie(sessionToken.name, token, { path: sessionToken.path });
  };`?

src/Conference/ConferenceDetails.jsx
```diff
@@ -56,7 +55,7 @@ const ConferenceDetails = ({ conference, id, isLoading, error }) => {
         <StyledDescription className="description">
           {description}
         </StyledDescription>
-        <Button onClick={deleteThisConference}>Delete Conference</Button>
+        <button onClick={deleteThisConference}>Delete Conference</button>
```
tonymaynard: not a fan of your own button?

src/StyledFormComponents.js
```diff
@@ -21,7 +21,7 @@ const StyledCard = styled.div`
   border-color: black;
   border-radius: 25px;
   background: #dbd8db;
-  width: 65%;
+  width: 900px;
```
tonymaynard: Whats the reasoning behind this?

src/api.js
```diff
@@ -84,3 +84,17 @@ export const loginUser = async user => {
...
   }
 };
+export const logoutUser = async token => {
+  try {
+    const response = await axios.delete(
+      "http://localhost:8080/auth/logout",
+      token
+    );
+    toast.success("Logout successful!");
+    return response.status;
```
tonymaynard: Out of curiosity do you expect to do anything this the status code? As I can't see you using it anywhere

src/App.js
```diff
@@ -21,15 +21,20 @@ const StyledApp = styled.div`
 
 function App() {
   const [cookies, setCookie, removeCookie] = useCookies(["sessionToken"]);
+  const cookieName = "sessionToken";
+  const cookieOptions = { path: "/" };
   const setToken = token => {
-    setCookie("sessionToken", token, { path: "/" });
+    setCookie(cookieName, token, cookieOptions);
```
tonymaynard: I'd be more explicit with the naming of the cookie, if someone else tries to add another cookie what would they call it?

src/Conference/ConferenceDetails.jsx
MerrettM: Our `Button` didn't really have any styling on it and we wanted to add styling to the logout button to make it fit in with the rest of the nav bar, so we decided it was easier to use the default button for the delete button and then style the custom one instead of adding functionality to change the styling of the `Button` by passing props. We could definitely add that functionality in if that'd be a better solution?

src/StyledFormComponents.js
MerrettM: We did this as a fix for an issue we were having with the text and buttons overflowing out of the card when it was resized.

src/api.js
MerrettM: I don't think we use it anywhere at the moment and I can't think of a reason to use it later on, should we remove this for now as it would be very easy to add back in if we do end up needing it?

src/Conference/ConferenceDetails.jsx
RaMorey: If the reusable button is not reusable, maybe it's time to re-think how it's been implemented! Reusable components should be able to have their styling adapted (For example we pass in a classname and use it to override the css)

MerrettM: Latest commit adds this functionality, much easier than we thought it would be! 😄

src/api.js
tonymaynard: If you're not using it, there's no need to really return that value :)

src/Button.jsx
```diff
@@ -2,13 +2,18 @@ import React from "react";
...
 
 export const StyledButton = styled.button`
-  text-align: center;
-  background: #322d38;
-  color: white;
-  border-color: #322d38;
-  font-size: 16px;
-  text-decoration: underline;
-  padding: 0;
+  ${props =>
```
tonymaynard: This is not really the best way to use styled components, all this is saying is if it has the prop `navLink` style this button if else, just use the standard button, are you saying you don't want styling at all?

MerrettM: Currently yes, the delete button didn't actually have any proper styling so as that is the only other case where we use the `Button` at the moment we left that blank with the idea that we could add in more conditionals like the one `navLink` uses as and when we need them.

src/App.js
```diff
@@ -21,15 +21,20 @@ const StyledApp = styled.div`
 
 function App() {
   const [cookies, setCookie, removeCookie] = useCookies(["sessionToken"]);
+  const cookieName = "sessionToken";
```
denishowew: Why not pass the cookieName variable to useCookies() instead of repeating the value?

src/Navbar.jsx
```diff
@@ -38,7 +51,18 @@ const Navbar = () => {
           <StyledLink to="/users/register">Register</StyledLink>
         </StyledListItem>
         <StyledListItem>
-          <StyledLink to="/users/login">Login</StyledLink>
+          {token ? (
+            <Button
+              navLink
+              onClick={() => {
```
denishowew: Would `onClick={ logout }` work?

src/App.js
longstaff-elsevier: make sure to declare the variable first too.

src/Button.jsx
tonymaynard: I don't think this is best practice but i think you'll have time at the end for stylist changes so I can talk you through maybe what to do then :)

src/Navbar.jsx
longstaff-elsevier: technically they are different signatures as there is an event passed with onClick, they are just ignoring it here. So that would work, but typescript might start complaining etc.
Generally I would say its preferable to buffer calls in this way unless there is a good reason.

PR #23, opened 2019-11-20
Gp19 86/87 profile page

sophiambolah:


src/Navbar.jsx
```diff
@@ -47,23 +47,32 @@ const Navbar = ({ token, deleteToken }) => {
...
-        </StyledListItem>
+        {token ? (
+          <>
+            <StyledListItem>
+              <StyledLink to={`/users/${token.userId}`}>Profile</StyledLink>
+            </StyledListItem>
+            <StyledListItem>
+              <Button
+                navLink
+                onClick={() => {
```
RaMorey: You should just be able to pass logout in here, instead of returning it in the arrow function

src/User/UserContainer.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+  useEffect(() => {
+    if (id) {
+      fetchData(id);
+    }
+  }, [id]);
+
+  return (
+    <Switch>
+      <Route path="/user/:id">
```
RaMorey: I'm slightly confused as you have the path to get here as /users but then in here it's changed to /user

src/api.js
```diff
@@ -72,6 +72,17 @@ export const loginUser = async user => {
     throw error;
   }
 };
+
+export const getUserById = async id => {
+  try {
```
RaMorey: If you're only throwing the error, you should be able to just remove the try catch all together and just do the logic inside the try

src/Navbar.jsx
MerrettM: This causes the logout function to trigger on every page refresh

RaMorey: You should be able to just pass the function in like this: `onClick={logout}`, then the button should be configured to use that function when it's clicked. Are you passing it in like this: `onClick={logout()}`?

src/api.js
MerrettM: We then get a "missing a catch or finally clause" error?

RaMorey: Updated my comment sorry

src/Navbar.jsx
```diff
@@ -30,7 +30,7 @@ const Navbar = ({ token, deleteToken }) => {
   const history = useHistory();
   const logout = async () => {
     try {
-      await logoutUser(token);
+      await logoutUser(token.token);
```
tonymaynard: should it be token.token? if so I think you need to rename some properties/variable

RaMorey: agreed, should it not be token.uuid?

MerrettM: Yeah, when we wrote the object in the backend it was written with naming like this:
`token = {userId: 1, token: "insert UUID String here"}`
so that is something we'd need the backend to change.

src/Navbar.test.js
```diff
@@ -23,7 +23,30 @@ afterEach(() => {
...
+    let wrapper;
+    act(() => {
+      wrapper = mount(
+        <Router history={history}>
+          <Navbar token={mockToken} />
+        </Router>
+      );
+    });
+
+    expect(wrapper.debug()).toMatchSnapshot();
```
tonymaynard: You may want to check something exists which is supposed to only when you are logged in as snapshots can easily be updated

src/Navbar.jsx
tonymaynard: In the meantime you can still rename the object, the backend doesn't assign that as its passed as a prop!

MerrettM: > In the meantime you can still rename the object, the backend doesn't assign that as its passed as a prop!

So rename it in `api` for now and then in `navBar` once they change it?

tonymaynard: currently in app you are passing the prop as `token` you can rename this there and subsequently rename the prop which `navBar` recieves this way it wont be `token.token`. Or similarly when you pass the setToken down to Login you could set the token and userId in state seperately. therefore for your `styledLink` you won't have to pass token.userId

src/Navbar.jsx
```diff
@@ -50,23 +50,30 @@ const Navbar = ({ sessionToken, deleteSessionToken }) => {
...
+              <StyledLink
+                className="profilePage"
+                to={`/users/${sessionToken.userId}`}
+              >
                 Profile
               </StyledLink>
             </StyledListItem>
             <StyledListItem>
-              <Button navLink onClick={logout}>
+              <Button classname="logout" navLink onClick={logout}>
```
dex-mendeley: Don't forget this is React, so should be className

MerrettM: Thanks that actually fixed an issue I had to find a workaround for in `Navbar.test`!

src/Navbar.jsx
```diff
@@ -47,23 +47,36 @@ const Navbar = ({ token, deleteToken }) => {
...
-              }}
-            >
-              Logout
-            </Button>
-          ) : (
-            <StyledLink to="/users/login">Login</StyledLink>
-          )}
-        </StyledListItem>
+        {sessionToken ? (
+          <>
```
j-hutchinson: These aren't needed as there is only one component `StyledListItem` in this block, everything is a child of it

j-hutchinson: likewise below

src/Navbar.test.js
```diff
@@ -23,8 +22,35 @@ afterEach(() => {
...
+    let wrapper;
+    act(() => {
+      wrapper = mount(
+        <Router history={history}>
+          <Navbar sessionToken={mockSessionToken} />
+        </Router>
+      );
+    });
+
+    expect(wrapper.debug()).toMatchSnapshot();
```
j-hutchinson: why are you checking the snapshot of the `debug()`? 
Why not just `expect(wrapper).toMatchSnapshot();`

src/api.js
```diff
@@ -2,32 +2,17 @@ import axios from "axios";
...
 };
 
 export const createNewConference = async conference => {
-  try {
-    await axios.post("http://localhost:8080/conferences", conference);
-  } catch (error) {
-    console.log(error);
-    throw error;
-  }
+  await axios.post("http://localhost:8080/conferences", conference);
```
j-hutchinson: Does this POST not return anything?

j-hutchinson: How will you know the call failed now?

src/Navbar.jsx
MerrettM: Are you talking about the `<></>`? There are two `StyledListItem` components within each set of them for whether the `sessionToken` cookie is or isn't present so we thought we did need them?

j-hutchinson: Yeah you're right, sorry i misread this. Didnt notice there were two `StyledListItems`

src/Navbar.test.js
MerrettM: We're using `debug()` here because it seems to obscure the `history` parameters which change for every time we run the tests and so the snapshot becomes useless when we don't use `debug()`.

src/api.js
MerrettM: It does but we don't use it for anything at them moment so that's why we deleted the return statement. For the sake of reducing future changes etc. would it be better to have the same `try{...} catch(error) {throw error}` sequence in each method and then do any logic like `toast` where these methods are called in the rest of the code?

PR #24, opened 2019-11-26
Gp19 94 restrict unauthorised actions

KaramKapoor:


src/Conference/AddConference.jsx
```diff
@@ -6,12 +6,12 @@ import {
   StyledCard
 } from "../StyledFormComponents";
 
-const AddConference = () => {
+const AddConference = sessionToken => {
```
tonymaynard: This arguement is actually `props` which you have just called sessionToken, do you not want to deconstruct it to `({ sessionToken })` ?

src/Conference/ConferenceDetails.jsx
```diff
@@ -48,15 +54,27 @@ const ConferenceDetails = ({ conference, id, isLoading, error }) => {
     <StyledForm>
       <StyledCard>
         <StyledCardHeading className="name">{name}</StyledCardHeading>
-        <StyledLink to={`/${id}/edit`}>Edit Conference</StyledLink>
+        {sessionToken ? (
```
tonymaynard: FYI you could write this as `{ sessionToken && ( <YourComponent /> )}` This works as the logical operator(&&) will actually evaluate and show the component if sessionToken exists.

src/Conference/AddConference.jsx
KaramKapoor: fixed!

src/Conference/ConferenceDetails.jsx
KaramKapoor: didn't realise that was possible - looks much cleaner now, thanks

src/Conference/AddConferenceForm.test.js
```diff
@@ -0,0 +1,110 @@
...
+      );
+    });
+
+    wrapper.find(AddConferenceForm).simulate("submit", ev);
+
+    await apiReturnValue;
+
+    expect(ev.preventDefault).toHaveBeenCalledTimes(1);
+    expect(createNewConference).toHaveBeenCalledTimes(1);
+
```
j-hutchinson: why the empty line?

src/Conference/AddConferenceForm.test.js
```diff
@@ -0,0 +1,110 @@
...
+    expect(history.location.pathname).toBe("/");
+  });
+  it("should stay on page if axios request unsucessful", async () => {
```
j-hutchinson: missing a `c` 😅

src/Conference/ConferenceDetails.jsx
```diff
@@ -18,7 +19,13 @@ export const StyledDescription = styled.p`
...
 
-const ConferenceDetails = ({ conference, id, isLoading, error }) => {
+const ConferenceDetails = ({
+  conference,
+  id,
+  isLoading,
+  error,
+  allCookies = {}
+}) => {
   const history = useHistory();
```
j-hutchinson: might be nice to make a variable here. `const sessionToken = allCookies.sessiontoken` as you use it in a couple places throughout the file

As the default is {}, youll have to do a check `allCookies ? allCookies.sessiontoken : null`

src/Conference/ConferenceDetails.test.js
```diff
@@ -27,8 +26,51 @@ describe("ConferenceDetails", () => {
...
+          <ConferenceDetails
+            conference={mockData}
+            id="1"
+            allCookies={mockCookie}
+          />
+        </Router>
+      );
+    });
+    wrapper.update();
+    console.log(wrapper.debug());
```
j-hutchinson: remove the logs

src/User/LoginForm.jsx
```diff
@@ -14,7 +16,12 @@ export const StyledLink = styled(Link)`
   padding: 10px;
 `;
 
-const LoginForm = ({ setSessionToken }) => {
+const LoginForm = () => {
+  const [, setCookie] = useCookies([cookieName]);
```
j-hutchinson: missing the first param here

src/Conference/ConferenceDetails.jsx
j-hutchinson: Likewise in other files as well

src/Conference/AddConferenceForm.test.js
```diff
@@ -0,0 +1,110 @@
...
+import { act } from "react-dom/test-utils";
+import { mount, configure } from "enzyme";
+import Adapter from "enzyme-adapter-react-16";
+import { createMemoryHistory } from "history";
+import { WrappedAddConferenceForm as AddConferenceForm } from "./AddConferenceForm";
+import { createNewConference } from "../api";
+
+configure({ adapter: new Adapter() });
+
+jest.mock("../api", () => ({
```
longstaff-elsevier: This would only work because the mock is hoisted, but you should be explicit about this being _before_ the import api (rather than importing the original and then mocking future imports)

src/Conference/AddConferenceForm.test.js
```diff
@@ -0,0 +1,110 @@
...
+  };
+
+  it("should send axios request", async () => {
+    expect.assertions(2);
+
+    const apiReturnValue = Promise.resolve(200);
+    createNewConference.mockImplementation(() => apiReturnValue);
+
+    let wrapper;
+    act(() => {
```
longstaff-elsevier: These shouldnt need the act wrapper?

act should only be used when there is active change in the state, usually to test mount events etc. rather than pure setup?

src/Conference/AddConferenceForm.jsx
```diff
@@ -82,8 +83,9 @@ const Form = () => {
...
         maxLength="20"
       />
-      <Input type="submit" value="submit" />
+      <Input type="submit" value="Submit" />
     </form>
   );
 };
-export default Form;
+export default withCookies(AddConferenceForm);
+export { AddConferenceForm as WrappedAddConferenceForm };
```
longstaff-elsevier: Specifically the default is wrapped with cookies HoC, here you are exporting the unwraped with the name wrapped?

src/Conference/AddConferenceForm.jsx
```diff
@@ -24,7 +25,7 @@ const Form = () => {
       dateTime: conference.dateTime + ":00Z"
     };
     try {
-      await createNewConference(newConference);
+      await createNewConference(newConference, allCookies.sessionToken.token);
```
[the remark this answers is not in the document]
longstaff-elsevier: I would remove cookies from the file completely and move into API (you might have to stop using hooks for that). This would save on complexity of components and centralise knowledge of the cookies into one place again, which you have lost in this refactor.

src/Conference/ConferenceDetails.jsx
```diff
@@ -48,17 +55,26 @@ const ConferenceDetails = ({ conference, id, isLoading, error }) => {
     <StyledForm>
       <StyledCard>
         <StyledCardHeading className="name">{name}</StyledCardHeading>
-        <StyledLink to={`/${id}/edit`}>Edit Conference</StyledLink>
+        {allCookies.sessionToken && (
```
longstaff-elsevier: Optimally you would look at abstracting this away from the cookie directly and into a util or provider of your own to make this more maintainable.

src/User/LoginForm.jsx
KaramKapoor: Leaving out the first param here and only having a comma just means no warnings are thrown on the front end, but it works the same way with [cookies, setCookie, removeCookie]

src/Conference/AddConferenceForm.jsx
KaramKapoor: Have changed the exports for AddConferenceForm and ConferenceDetails: prefixed the main one with the word 'Cookie', then exported as default; and exported the cookie-less one inline for test usage.

src/Conference/AddConferenceForm.jsx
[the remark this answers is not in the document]
KaramKapoor: I've taken a note of this, we'll bring it up in refinement and probably add it to a new ticket as it it out of scope for this branch

src/Conference/ConferenceDetails.jsx
KaramKapoor: We will be looking at implementing cookies in a cleaner way, as part of the solution addressed in the other comment

src/Conference/AddConferenceForm.jsx
```diff
@@ -1,10 +1,12 @@
...
-import { createNewConference } from "./api";
 import { useHistory } from "react-router-dom";
+import { withCookies } from "react-cookie";
+import Input from "../Input";
+import { createNewConference } from "../api";
 
-const Form = () => {
+export const AddConferenceForm = ({ allCookies = {} }) => {
   let history = useHistory();
+  const token = allCookies.sessionToken.token;
```
j-hutchinson: if `allCookies` is equal to your default, this will break as you can read property token from undefined. 
`allCookies.sessionToken ? allCookies.sessionToken : null` is one work around

src/Conference/ConferenceDetails.jsx
```diff
@@ -18,8 +19,15 @@ export const StyledDescription = styled.p`
...
-const ConferenceDetails = ({ conference, id, isLoading, error }) => {
+export const ConferenceDetails = ({
+  conference,
+  id,
+  isLoading,
+  error,
+  allCookies = {}
+}) => {
   const history = useHistory();
+  const sessionToken = allCookies.sessionToken;
```
j-hutchinson: Unlike my previous comment, this is okay as sessionToken will be `undefined` here opposed to an error

src/Conference/ConferenceDetails.test.js
```diff
@@ -27,8 +26,50 @@ describe("ConferenceDetails", () => {
...
+          <ConferenceDetails
+            conference={mockData}
+            id="1"
+            allCookies={mockCookie}
+          />
+        </Router>
+      );
+    });
+    wrapper.update();
+    expect(wrapper.find(".name").get(0).props.children).toBe(mockData.name);
```
j-hutchinson: at this point you should make a shorthand for this.
`const  findElement = (identifier: string) => wrapper.find(identifier).get(0).props.children)`
then
`expect(findElement('name')).toBe(mockData.name);`
If you then use this in multiple places you can make it a util and import in your tests

(I've not tested this and just wrote it here, may not be perfect 😅)

src/Navbar.jsx
```diff
@@ -26,33 +28,39 @@ export const StyledLink = styled(Link)`
...
     <StyledNavbar>
       <StyledList>
         <StyledListItem>
           <StyledLink to="/">Home</StyledLink>
         </StyledListItem>
-        <StyledListItem>
-          <StyledLink to="/add">Add Conference</StyledLink>
-        </StyledListItem>
-        {sessionToken ? (
+        {allCookies.sessionToken ? (
```
j-hutchinson: for consistency across components you should make this a variable as well.

src/Navbar.jsx
```diff
@@ -82,4 +90,5 @@ const Navbar = ({ sessionToken, deleteSessionToken }) => {
   );
 };
 
-export default Navbar;
+export default withCookies(Navbar);
+export { Navbar as WrappedNavbar };
```
j-hutchinson: Do you need to export this as `WrapperNavBar`? Why not just export the component at line 31

src/Conference/AddConferenceForm.jsx
KaramKapoor: OK just replaced it with `const token = allCookies.sessionToken ? allCookies.sessionToken.token : null` - it's still working so hope that's alright

src/Navbar.test.js
```diff
@@ -43,10 +43,13 @@ describe("Navbar", () => {
     expect(wrapper.find(".login").length).toBe(0);
     expect(wrapper.find(".profilePage").get(0).props.children).toBe("Profile");
     expect(wrapper.find(".logout").get(0).props.children).toBe("Logout");
```
j-hutchinson: If you create a util file for the `findElement` function you created, you can import it and reuse it here as well

KaramKapoor: Took a note of this, will bring this up as a ticket and address these changes during our technical debt stage

src/App.js
```diff
@@ -4,10 +4,10 @@ import { Switch, Route } from "react-router-dom";
 import ConferenceContainer from "./Conference/ConferenceContainer";
 import ConferenceList from "./Conference/ConferenceList";
 import AddConference from "./Conference/AddConference";
-import Navbar from "./Navbar";
+import CookieNavbar from "./Navbar";
```
longstaff-elsevier: This doesnt need to change name here? just use `Navbar`

PR #25, opened 2019-11-26
GP19-90 Add edit button

karenzamani:


src/App.js
```diff
@@ -46,6 +47,9 @@ function App() {
             <Route path="/add">
               <AddConference />
             </Route>
+            <Route path="/users/edit/:id">
```
tonymaynard: Should this not be in UserContainer?

src/User/UpdateProfile.jsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import {
+  StyledForm,
+  StyledCard
+} from "../StyledFormComponents";
+
+const UpdateProfile = ({  }) => {
```
tonymaynard: the {} are not needed if you haven't got any props to destructor

src/User/ProfilePage.jsx
```diff
@@ -14,19 +29,19 @@ const ProfilePage = ({ user, isLoading, error }) => {
...
-        <StyledCardHeading className="name">
-          Hello {firstName} {lastName}
-        </StyledCardHeading>
-        <p className="username">Username: {username}</p>
-        <p className="email">Email: {email}</p>
-        <p className="occupation">Occupation: {occupation}</p>
-      </StyledCard>
-    </StyledForm>
+    <StyledCard>
+      <StyledLink className="editLink" to={`/users/edit/${id}`}>Edit</StyledLink>
```
tonymaynard: You may want to check if the current users Id matches the person you are viewing in order to show this button

src/User/ProfilePage.test.jsx
```diff
@@ -0,0 +1,50 @@
...
+        occupation: "Fun" 
+  };
+
+  it("renders profile information", async () => {
+    expect.assertions(5);
+
+    const history = createMemoryHistory();
+    history.push("/users/11");
+
+    let wrapper;
```
tonymaynard: Doesn't need to be let as we aren't reassigning, const should be fine

PR #27, opened 2019-12-02
GP19-117: add ability to select manager on registration

MerrettM:
Hopefully this will be switched to a searchable box in the future so this is somewhat a placeholder.

### DIFF
--- a/src/User/RegistrationForm.jsx
+++ b/src/User/RegistrationForm.jsx
@@ -1,7 +1,7 @@
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
 import { useHistory } from "react-router-dom";
 import Input from "../Input";
-import { createNewUser } from "../api";
+import { createNewUser, getUserList } from "../api";
 import {
   StyledCardHeading,
   StyledForm,
@@ -15,9 +15,24 @@ const RegistrationForm = () => {
     lastName: "",
     email: "",
     occupation: "",
+    managerId: "",
     username: "",
     password: ""
   });
+  const [managers, setManagers] = useState([]);
+
+  const fetchData = async () => {
+    try {
+      const userList = await getUserList();
+      setManagers(userList);
+    } catch (error) {
+      console.log(error);
+    }
+  };
+
+  useEffect(() => {
+    fetchData();
+  }, []);
 
   const handleChange = event => {
     const { name, value } = event.target;
@@ -36,6 +51,7 @@ const RegistrationForm = () => {
       console.log(error);
     }
   };
+
   const onSubmit = ev => {
     ev.preventDefault();
     submitForm();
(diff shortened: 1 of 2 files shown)


src/User/RegistrationForm.jsx
```diff
@@ -15,9 +15,24 @@ const RegistrationForm = () => {
...
     occupation: "",
+    managerId: "",
     username: "",
     password: ""
   });
+  const [managers, setManagers] = useState([]);
+
+  const fetchData = async () => {
+    try {
+      const userList = await getUserList();
```
tonymaynard: Why are managers all users?

MerrettM: To implement manager profiles the BE users table is being altered to include a `managerId` column which connects each user to another user's `id`. Therefore managers are users which are assigned to their role in the system by users selecting their name when they register. So if a manager has registered with the app but none of their employees have registered yet they wouldn't be searchable as a manager despite being one. So essentially we have to use get all users in this situation as otherwise some managers might be missed out.

Apologies for the wall of text 😄 let me know if it doesn't make any sense!

RaMorey: Yeh I'm a bit confused on what this PR is for, its a list of users?

MerrettM: I've renamed the PR, since it took the commit message as it's name, which wasn't very descriptive.

Essentially this PR is for adding in a `<select>` on the registration page which allows you to select your manager from a list of users.

MerrettM: And the associated API request

tonymaynard: So am I right in assuming if this is deployed and we have 1000's of users the options will be endless?

MerrettM: If it's left in the current state yes, when we discussed this in Sprint Planning we decided to do it this way as a placeholder and then in the future add something like autocomplete/search functionality. I think @GBurley-Jones and @sophiambolah will probably be better at answering questions about that.

RaMorey: So, from what I've gathered, it sounds like the ticket shouldn't really be done yet. What you've done is fine, but just something to think about going forward is that it's generally not worth implementing something as a placeholder until its actually ready to go. There's several reasons for this but one of them can be that product might get the impression that it's working and keep it in, or you might get busy with something else and not come back to it.

I've spoken to a few people about how you're implementing this feature and sounds like you want to use elastic search for something that should be a simple SQL to check if your database has a certain string in it for the names of the users? 

Just something you might want to bring up with the team - think elastic search is major overkill for a search on user name

src/User/RegistrationForm.jsx
```diff
@@ -83,6 +99,15 @@ const RegistrationForm = () => {
             required
             maxLength="100"
           />
+          <label>Manager: </label>
+          <select onChange={handleChange} name="managerId" defaultValue="">
+            <option value="">---- select ----</option>
+            {managers.map(manager => (
```
RaMorey: Also this should probably be users.map, because they're not managers at this point - it's just a list of users that you select to choose your manager from

MerrettM: yeah I changed it to managers and manager because we already use user on that page, I'll find a better way to name them when this PR reopens.

PR #28, opened 2019-12-03
Gp19 91 Prefilled edit form

karenzamani:


src/User/ProfilePage.jsx
```diff
@@ -32,16 +34,18 @@ const ProfilePage = ({ error, isLoading, user }) => {
...
-      <StyledLink className="editLink" to={`/users/edit/${id}`}>Edit</StyledLink>        
+      <StyledLink className="editLink" to={`/users/${id}/edit`}>Edit</StyledLink>        
       <StyledCardHeading className="name">
         {`Hello, ${firstName} ${lastName}!`}
       </StyledCardHeading>
       <p className="username">{`Username: ${username}`}</p>
       <p className="email">{`Email: ${email}`}</p>
       <p className="occupation">{`Occupation: ${occupation}`}</p>
-    </StyledCard>
+      </StyledCard>
```
RaMorey: Could probably just format this slightly nicer for readability. Shift card in 1 tab and Styled profile back one so it's indented once per level

src/User/UpdateProfile.jsx
```diff
@@ -1,17 +1,82 @@
...
+        setUpdatedUser({ ...updatedUser, [name]: value });
+    }
+    
+    const submitForm = async () => {
+        try {
+          await updateUserById(id, updatedUser, sessionToken.token);
+          setUser(updatedUser);
+          history.push(`/users/${id}`);
+        } catch (error) {
+          console.log(error);
```
RaMorey: If you're only logging the error you can get rid of the try catch

src/User/UpdateProfile.jsx
```diff
@@ -1,17 +1,82 @@
...
+        }
+    };
+
+    return (
+        <StyledForm>
```
RaMorey: I think you need to fix your prettier settings, the indentation and formatting is quite inconsistent across different files

src/User/UserContainer.jsx
```diff
@@ -14,27 +15,44 @@ const UserContainer = ({ token }) => {
     setIsLoading(true);
 
     try {
-      const user = await getUserById(userId, token);
+      const user = await getUserById(userId, sessionToken.token);
       setUser(user);
+      setUpdatedUser({
```
RaMorey: Would you be able to explain what this line and the one above it are doing? I'm a little confused

src/User/UserContainer.jsx
```diff
@@ -14,27 +15,44 @@ const UserContainer = ({ token }) => {
...
+        ...user
+      });
     } catch (error) {
       setError(true);
     }
     setIsLoading(false);
   };
 
   useEffect(() => {
+    console.log(id);
```
RaMorey: Can remove this now

src/User/UserContainer.jsx
RaMorey: I think after reading it a bit more, you should consider moving the updatedUser state into the UpdateProfile component

src/api.js
```diff
@@ -61,9 +61,11 @@ export const loginUser = async user => {
 };
 
 export const getUserById = async (id, token) => {
+  console.log(token)
```
RaMorey: log log

RaMorey:
Also you probably want to add in some tests for the updateProfile component

src/api.js
karenzamani: log

PR #29, opened 2019-12-03
Gp19 117 select manager search box and dropdown

MerrettM:


src/api.js
```diff
@@ -48,6 +48,13 @@ export const createNewUser = async user => {
   }
 };
 
+export const getSearchResults = async input => {
+  const response = await axios.get(
+    `http://localhost:8080/users/search/${input}`
```
RaMorey: You would probably want to pass the input in as a query parameter, not as an id

MerrettM: Do you mean pass it in the request body not as part of the url?

RaMorey: No, have a read of query strings https://stackoverflow.com/questions/39266970/what-is-the-difference-between-url-parameters-and-query-strings

src/User/RegistrationForm.jsx
```diff
@@ -1,7 +1,8 @@
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
 import { useHistory } from "react-router-dom";
 import Input from "../Input";
-import { createNewUser } from "../api";
+import { createNewUser, getSearchResults } from "../api";
+import { DebounceInput } from "react-debounce-input";
```
j-hutchinson: this would have been helpful for SUTD a while back 😂

MerrettM: @RaMorey's idea!

src/api.js
```diff
@@ -48,6 +48,13 @@ export const createNewUser = async user => {
   }
 };
 
+export const getSearchResults = async input => {
+  const response = await axios.get(
+    `http://localhost:8080/users/search?query=${input}`
```
RaMorey: I don't think you need the search bit. You want the end point to be exactly the same as the normal GET list endpoint, but with an extra querystring to filter it down

RaMorey: Again, that would be a back end change really - so depends if they approve the back end or not aha

MerrettM: Yeah this has been commented in the BE PR as something to think about but not neccesarily something that needs to be done at the moment. Karam has decided that if we do change this it will probably be after the feature freeze.

src/User/RegistrationForm.jsx
```diff
@@ -15,9 +16,31 @@ const RegistrationForm = () => {
     lastName: "",
     email: "",
     occupation: "",
+    managerId: "",
```
tonymaynard: If the Id is a Number the default state probably should be `null` and not an empty string

PR #31, opened 2019-12-05
Gp19 119 manager profile link

MerrettM:
Adds the ability to look at your manager's profile as well as your own.


src/User/ProfilePage.jsx
```diff
@@ -29,17 +38,47 @@ const ProfilePage = ({ error, isLoading, user }) => {
...
+          Edit
+        </StyledEditLink>
+      )}
       <StyledCardHeading className="name">
         {`Hello, ${firstName} ${lastName}!`}
       </StyledCardHeading>
       <p className="username">{`Username: ${username}`}</p>
       <p className="email">{`Email: ${email}`}</p>
       <p className="occupation">{`Occupation: ${occupation}`}</p>
+      {managerName && displayManager()}
```
j-hutchinson: Should look into useMemo hook here. This function is going to be computed every time the component is rendered even though the value wont change. If you memoize the function, the value will be cached and only recomputed when a dependancy is updated (ie managerName)

j-hutchinson: https://reactjs.org/docs/hooks-reference.html#usememo

src/User/ProfilePage.test.jsx
```diff
@@ -16,32 +16,134 @@ jest.mock("react-router-dom", () => {
...
+        />
+      </Router>
+    );
+
+    wrapper.update();
+    expect(wrapper.find(".editLink").get(0).props.to).toBe(
+      `/users/${mockDataWithManager.id}/edit`
+    );
+    expect(findElement(".name")).toBe(
+      `Hello, ${mockDataWithManager.firstName} ${mockDataWithManager.lastName}!`
```
j-hutchinson: You can expect this to be `Hello, Ross Morey` i dont think you really need to read from the mock here like this. But its more a personal opinion on testing
We do it a bit of both ways in SUTD

src/User/ProfilePage.jsx
```diff
@@ -29,17 +38,47 @@ const ProfilePage = ({ error, isLoading, user }) => {
     return <p>An error has occurred...</p>;
   }
 
-  const { id, username, firstName, lastName, email, occupation } = user || {};
+  const { id, username, firstName, lastName, email, occupation, managerId } =
+    user || {};
+
+  const displayManager = () => {
+    if (String(managerName) === "None Assigned") {
```
j-hutchinson: what type is managerName, shouldn't it already be a string?

src/User/ProfilePage.test.jsx
```diff
@@ -16,32 +16,134 @@ jest.mock("react-router-dom", () => {
...
+  };
+
+  const mockManagerName = "Joe Bloggs";
+
+  const mockNoManagerName = "None Assigned";
+
+  let wrapper;
+  const findElement = identifier => {
+    return wrapper.find(identifier).get(0).props.children;
+  };
```
j-hutchinson: This should be moved to a util file now, this is being used in a couple places

src/User/ProfilePage.test.jsx
```diff
@@ -16,32 +16,134 @@ jest.mock("react-router-dom", () => {
...
+      `Username: ${mockDataWithoutManager.username}`
+    );
+    expect(findElement(".email")).toBe(
+      `Email: ${mockDataWithoutManager.email}`
+    );
+    expect(findElement(".occupation")).toBe(
+      `Occupation: ${mockDataWithoutManager.occupation}`
+    );
+    expect(findElement(".manager")).toBe(`Manager: ${mockNoManagerName}`);
+    expect(wrapper.find(".managerLink").length).toBe(0);
```
j-hutchinson: If you use a snapshot test, you're probably safe just checking the that child contains `Ross Morey` for example rather than the more verbose `Hello, ...`
If you correctly use snapshots and look at what is snapshotted, you will be able to tell if Hello isn't present or suddenly gets removed during future work

src/User/ProfilePage.test.jsx
MerrettM: I left it like this cause I figured it'd stop the test breaking in the future if someone changed the mock data.

src/User/UserContainer.jsx
```diff
@@ -10,35 +10,72 @@ const UserContainer = () => {
...
+        const manager = await getUserById(
+          managerId,
+          cookies.sessionToken.token
+        );
+        setManagerName(`${manager.firstName} ${manager.lastName}`);
+      } catch (error) {
+        setError(true);
+      }
+    } else {
+      setManagerName("None Assigned");
```
longstaff-elsevier: This should probably be a constant that can be read in other places, or as a generic label that can be overwritten in the point of display

src/User/ProfilePage.jsx
longstaff-elsevier: This should probably be a constant imported/null etc. so you cant fall out of sync

MerrettM: It's a string, I was getting an error at one point that required me to do that `String()` thing but having just removed it I'm not anymore...

j-hutchinson: I'm sad that you guys didn't use Typescript 😔

src/User/UserContainer.jsx
```diff
@@ -10,35 +10,72 @@ const UserContainer = () => {
   const [user, setUser] = useState(null);
   const [isLoading, setIsLoading] = useState(true);
   const [error, setError] = useState(false);
+  const [isCurrentUser, setIsCurrentUser] = useState(false);
```
longstaff-elsevier: This should probably be derived data, not set in state

src/User/UserContainer.jsx
```diff
@@ -10,35 +10,72 @@ const UserContainer = () => {
...
+
     setIsLoading(false);
   };
 
   useEffect(() => {
     if (id) {
       fetchData(id);
     }
+    if (isCurrentUser) {
+      getManagerName(user.managerId);
```
longstaff-elsevier: this also therefore relies on user.managerId, should probably be a separate useEffect?

src/User/UserContainer.jsx
```diff
@@ -10,35 +10,72 @@ const UserContainer = () => {
...
+    }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [id]);
+  }, [id, isCurrentUser]);
 
   return (
     <Switch>
-      <Route path="/users/edit/:id">
+      <Route path="/users/:id/edit">
```
tonymaynard: Is it possible to navigate to editing another users page? i.e. your id is `1` and you hit `/users/5/edit`. If so you may want to have a redirect condition now you have the ability to check whether the path belongs to current user

src/User/UserContainer.jsx
```diff
@@ -10,35 +10,72 @@ const UserContainer = () => {
...
         <UpdateProfile />
       </Route>
       <Route path="/users/:id">
-        <ProfilePage user={user} isLoading={isLoading} error={error} />
+        <ProfilePage
+          user={user}
+          isLoading={isLoading}
+          error={error}
+          isCurrentUser={isCurrentUser}
```
longstaff-elsevier: this should be checked in the component, `user.id === path.id`?

src/User/UserContainer.jsx
MerrettM: Thats a good point, you should only be able to edit your own details.

src/User/UserContainer.jsx
MerrettM: Now that we have the redirect should this still be checked in the component or in the container?

tonymaynard: In the component you are using it to check whether to render the edit link button, therefore some kind of check is still needed so the user doesn't click on an edit button when its not their profile

src/User/UserContainer.jsx
```diff
@@ -64,7 +64,7 @@ const UserContainer = () => {
   return (
     <Switch>
       <Route path="/users/:id/edit">
-        <UpdateProfile />
+        {isCurrentUser ? <UpdateProfile /> : <Redirect to="/" />}
```
tonymaynard: I would probably suggest redirecting to their profile instead of sending them to the home page

src/User/UserContainer.jsx
MerrettM: Ah yeah don't worry we still have that, I was more asking abut if it should be done in situ or not. I've made a function to return the true or false which I call in the conditional rendering checks.

src/TestUtils.js
```diff
@@ -0,0 +1,9 @@
+export let wrapper;
+
+export const setWrapper = componantToBeWrapped => {
```
shanassaf: *component 😅

src/TestUtils.js
```diff
@@ -0,0 +1,9 @@
...
+  wrapper = componantToBeWrapped;
```
shanassaf: and here

src/User/UserContainer.jsx
```diff
@@ -1,53 +1,96 @@
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [id]);
+
+  useEffect(() => {
+    if (isCurrentUser) {
+      getManagerName(user);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isCurrentUser, user]);
```
longstaff-elsevier: im not sure that this will work, the useEffect triggers when the _value_ changes, this will basically always check if `function isCurrentUser` === `function isCurrentUser`, which is probably always false as you are redefining it?

you might be able to do
```
useEffect(() => {
    ...
  }, [isCurrentUser(), user]);
```

The fact that you are having to use lint disable is a hint you are doing something wrong!

src/User/UserContainer.jsx
```diff
@@ -1,53 +1,96 @@
...
+
+  useEffect(() => {
+    if (id) {
+      fetchData(id);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [id]);
+
+  useEffect(() => {
+    if (isCurrentUser) {
```
longstaff-elsevier: isCurrentUser is the function, so this is always true

src/User/ProfilePage.jsx
```diff
@@ -31,21 +41,44 @@ const ProfilePage = ({ error, isLoading, user }) => {
     return <p>An error has occurred...</p>;
   }
 
-  const { id, username, firstName, lastName, email, occupation } = user || {};
+  const { id, username, firstName, lastName, email, occupation, managerId } =
+    user || {};
+
+  const displayManager = () => {
+    if (managerName === "No_Manager") {
```
longstaff-elsevier: good start, for maintainability this should probably be a universal constant that you are importing instead of redeclared in several places, make one centralised place for this value to live and import it to other places that need it.

src/User/UserContainer.jsx
MerrettM: I think we fixed the triggering issues but without the lint I get three warnings now:
`Line 67:6:   React Hook useEffect has missing dependencies: 'getManagerName', 'isCurrentUser', and 'user'. Either include them or remove the dependency array  react-hooks/exhaustive-deps`
` Line 67:7:   React Hook useEffect has a complex expression in the dependency array. Extract it to a separate variable so it can be statically checked          react-hooks/exhaustive-deps`
`Line 67:24:  React Hook useEffect has a complex expression in the dependency array. Extract it to a separate variable so it can be statically checked          react-hooks/exhaustive-deps`
I can't really see how to fix these as they seem to be similar warnings as the ones we get for the fetch data useEffect and every time I try to do what it asks it breaks something else?

MerrettM: I tried using the `useCallback` solution but that doesn't actually deal with the `getManagerName` missing dependencies thing, it actually doesn't seem to be doing anything different really.

src/App.js
```diff
@@ -4,63 +4,51 @@ import { Switch, Route } from "react-router-dom";
...
-				</StyledApp>
-				<ToastContainer position="bottom-right" autoClose={5000} pauseOnHover />
-			</BrowserRouter>
-		</CookiesProvider>
-	);
+  return (
+    <CookiesProvider>
+      <BrowserRouter>
+        <StyledApp>
+          <CookieNavbar />
```
tonymaynard: Wouldn't prefix with `Cookie` App doesn't care whether the NavBar needs a cookie or not right

src/Conference/ConferenceDetails.test.js
```diff
@@ -33,26 +34,23 @@ describe("ConferenceDetails", () => {
...
     act(() => {
-      wrapper = mount(
-        <Router history={history}>
-          <ConferenceDetails
-            conference={mockData}
-            id="1"
-            allCookies={mockCookie}
-          />
-        </Router>
+      setWrapper(
```
tonymaynard: Could go one further and setMountedWrapper, then mount the components in the util

src/User/ProfilePage.test.jsx
```diff
@@ -16,32 +17,135 @@ jest.mock("react-router-dom", () => {
...
+    firstName: "Ross",
+    lastName: "Morey",
+    email: "Ross.Morey@live.co.uk",
+    occupation: "Fun",
+    managerId: null
+  };
+
+  const mockManagerName = "Joe Bloggs";
+
+  const mockNoManagerName = "No_Manager";
```
tonymaynard: Could import this from your constants so if this changes the test won't need an additional change

src/User/UserContainer.jsx
```diff
@@ -1,53 +1,99 @@
...
+  }, [id]);
+
+  useEffect(() => {
+    if (isCurrentUser()) {
+      getManagerName(user);
+    } else {
+      setManagerName("");
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isCurrentUser(), user ? user.id : null]);
```
tonymaynard: might be better to assign this to a constant ` const userId = user ? user.id : null`

src/User/UserContainer.jsx
```diff
@@ -1,53 +1,99 @@
...
+          <UpdateProfile
+            user={user}
+            id={id}
+            setUser={setUser}
+            isLoading={isLoading}
+            error={error}
+            sessionToken={cookies.sessionToken}
+          />
+        ) : (
+          <Redirect to={`/users/${cookies.sessionToken.userId}`} />
```
tonymaynard: what happens if a user isn't logged in? probably best to just redirect to `"/users/:id` right

MerrettM: You aren't allowed to see `/users/:id` if you aren't logged in though so possibly that needs to be redirected in app not user container

MerrettM: I've added an extra layer of redirects which trigger when you're not logged in taking you away from the prohibited route to `/users/login` so then the other redirect can only ever be triggered when you are logged in and so `/users/${cookies.sessionToken.userId}` is a valid route.

tonymaynard: My impression was when you are viewing a user who is your employee you would still use this route of `user\:id`? It doesn't matter for now was just bringing to your attention

MerrettM: That is the case, the redirect in `App` stops you viewing the `/user/:id` route if you aren't logged in, the redirect in `UserContainer` stops you viewing the `/user/:id/edit` unless you are the user being edited. So if you're a manager who is logged in you can still see your employees, its only if you aren't logged in that you can't see the `/users/:id` route.

src/User/UserContainer.jsx
MerrettM: I'm going to merge this now (since this technically does work and it's a warning not an error which looks like it might take a while to sort out) because we need to get the feature out before our demo, but I have written down the issue and will make it into it's own ticket to be done post feature freeze.

PR #32, opened 2019-12-06
GP19-133 Scraping Page with text box and submit

karenzamani:


src/App.js
```diff
@@ -33,6 +34,9 @@ function App() {
             <Route path="/add">
               <AddConference />
             </Route>
+            <Route path="/addURL">
```
RaMorey: whats the reason behind not putting it in the add conference component?

src/StyledFormComponents.js
```diff
@@ -29,6 +29,7 @@ const StyledCard = styled.div`
   }
   vertical-align: middle;
   margin: 0 auto;
+  margin-top: 1px;
```
RaMorey: Is this required

karenzamani: Urm I thought it was because the styled card was displaying funny in my browser but Ive double checked it and it makes no difference.
I'll remove it

src/App.js
karenzamani: Hmmm, I guess they could go on the same page. Theres no reason, I just hadn't considered it.

RaMorey: Okay just wondering if there was a reason for it! Makes the urls more readable if they're in the same place :)

RaMorey: I'll approve you can decide how you want the structure to be :)

src/User/UserContainer.jsx
```diff
@@ -2,46 +2,52 @@ import React, { useState, useEffect } from "react";
...
-  const [isLoading, setIsLoading] = useState(true);
-  const [error, setError] = useState(false);
-  const { id } = useParams();
-  const [cookies] = useCookies([cookieName]);
+const UserContainer = ({ sessionToken }) => {
+	const [user, setUser] = useState(null);
+	const [isLoading, setIsLoading] = useState(true);
+	const [error, setError] = useState(false);
+	const { id } = useParams();
+	const [cookies] = useCookies([cookieName]);
```
tonymaynard: you're using hooks and a higher order component at the bottom too, the higher ordered component will inject `userContainer` with the props allCookies & cookies where as the hooks access the cookies directly. if you're going with the hook approach please remove the withCookies around the export, and you don't need to pass in `sessionToken` as props as you have access to cookies here so it'll be the same as `cookies.sessionToken`

src/Webscrape/WebscrapePage.jsx
```diff
@@ -0,0 +1,36 @@
+import React, {useState} from "react";
+import Input from "../Input";
+import { submitNewURL } from "../api";
+import { withCookies } from "react-cookie";
+import { useHistory } from "react-router-dom";
+
+const WebscrapePage = ({ allCookies = {} }) => {
```
tonymaynard: if you're using `allCookies` prop you need to wrap the export default with the `withCookies`

src/Webscrape/WebscrapePage.jsx
```diff
@@ -0,0 +1,36 @@
...
+	let history = useHistory();
+	const token = allCookies.sessionToken ? allCookies.sessionToken.token : null;
+	const [URL, setURL] = useState({
```
tonymaynard: any reason this isn't just 
```
const [URL, setURL] = useState("");
``` ?

File: src/User/UserContainer.jsx (hunk 3 above)
Comment (karenzamani) [reply src/User/UserContainer.jsx]: Hi Tony, are you free at all to go through this today? I have some questions

Comment (tonymaynard) [reply src/User/UserContainer.jsx]: Sure, have a read of this might help explain the difference https://medium.com/javascript-scene/do-react-hooks-replace-higher-order-components-hocs-7ae4a08b7b58

File: src/User/UserContainer.jsx [hunk 6]
```diff
@@ -1,47 +1,53 @@
...
-    try {
-      const user = await getUserById(userId, cookies.sessionToken.token);
-      setUser(user);
-    } catch (error) {
-      setError(true);
-    }
-    setIsLoading(false);
-  };
+		try {
+			const user = await getUserById(id, cookies.sessionToken.token);
```
tonymaynard: Is there any point where this sessionToken doesn't exist, Can you get to this page when logged out i can't remember?

src/Webscrape/WebscrapePage.jsx
```diff
@@ -0,0 +1,45 @@
...
+				type="text"
+				name="URL"
+				value={URL}
+				onChange={handleChange}
+			/>
+			<Input type="submit" value="Submit" />
+		</form>
+	);
+};
+const CookiesWebscrapePage = withCookies(WebscrapePage);
```
tonymaynard: can just be `export default withCookies(WebscrapePage)`

src/Webscrape/WebscrapePage.jsx
```diff
@@ -0,0 +1,45 @@
...
+	const token = allCookies.sessionToken ? allCookies.sessionToken.token : null;
+	const [URL, setURL] = useState("");
+
+	const handleChange = event => {
+		setURL(event.target.value);
+	};
+
+	const submitURL = async () => {
+		try {
+			await submitNewURL(URL, token);
```
tonymaynard: may want to check whether the url exists before sending this request

karenzamani: urmmm, like with an if statement?

src/User/UserContainer.jsx
karenzamani: Only logged in members should be able to get to this page!

src/Webscrape/WebscrapePage.jsx
karenzamani: or using something similar to the above comment? getUserById(id, cookies.sessionToken.token);

karenzamani: this has been added in!

PR #34, opened 2019-12-17
GP19-159: Redirect Conditional Re-Render Bug Fix

MerrettM:
fixing a bug that caused the setting of a new cookie not to trigger a re-render of the `Component`/`Redirect` conditionals in `App`


src/Conference/AddConferenceRedirector.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+  return (
+    <>
+      {cookies.sessionToken ? (
+        <AddConference />
+      ) : (
+        <Redirect to="/users/login" />
+      )}
+    </>
+  );
```
longstaff-elsevier: should be able to be simplified to 

`return cookies.sessionToken ? <AddConference /> : <Redirect to="/users/login" />`

?

src/User/UserContainerRedirector.jsx
```diff
@@ -0,0 +1,21 @@
...
+  return (
+    <>
+      {cookies.sessionToken ? (
+        <UserContainer />
+      ) : (
+        <Redirect to="/users/login" />
+      )}
+    </>
+  );
+};
```
longstaff-elsevier: make generic?

`
const UserRedirector = ({redirectPath, children}) => {
  const [cookies] = useCookies([cookieName]);

  return (
      cookies.sessionToken ? children : <Redirect to={redirectPath} />
`

src/App.js
```diff
@@ -21,8 +20,6 @@ const StyledApp = styled.div`
   min-height: 100vh;
 `;
 function App() {
-  const [cookies] = useCookies([cookieName]);
```
longstaff-elsevier: Original bug is probably that you are referencing this outside of the provider (first line of render) so wont hook into updates

src/App.js
```diff
@@ -39,7 +42,9 @@ function App() {
               <LoginForm />
             </Route>
             <Route path="/users/:id">
-              <UserContainerRedirector />
+              <CookieOrRedirect redirectPath="/users/login">
```
tonymaynard: May I ask why you chose to prefix with Cookie again? A user could have multiple cookies on their browser

tonymaynard: Perhaps AuthRedirect ?

MerrettM: Thats true, I used that one mainly because it was descriptive but AuthRedirect would probably be more accurate to it's function

PR #35, opened 2019-12-19
Gp19 123 list employees

MerrettM:
User's should be able to see a list of their employees when they are on their own profiles

### DIFF
--- a/src/Navbar.test.js
+++ b/src/Navbar.test.js
@@ -8,9 +8,10 @@ jest.mock("react-router-dom", () => {
 import React from "react";
 import { MemoryRouter } from "react-router-dom";
 import { act } from "react-dom/test-utils";
-import { configure, mount } from "enzyme";
+import { configure } from "enzyme";
 import Adapter from "enzyme-adapter-react-16";
 import { Navbar } from "./Navbar";
+import { wrapper, setMountedWrapper, findElement } from "./TestUtils";
 
 configure({ adapter: new Adapter() });
 
@@ -29,9 +30,8 @@ describe("Navbar", () => {
   it("renders a logged-in navbar with text", () => {
     expect.assertions(6);
 
-    let wrapper;
     act(() => {
-      wrapper = mount(
+      setMountedWrapper(
         <MemoryRouter initialEntries={[{ pathname: "/", key: "testKey" }]}>
           <Navbar allCookies={mockCookie} />
         </MemoryRouter>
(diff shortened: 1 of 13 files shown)


src/User/EmployeeList.jsx
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import styled from "styled-components";
+import { Link } from "react-router-dom";
+
+const StyledEmployeesList = styled.ul`
```
tonymaynard: I Think this is exactly the same as another ul/li combination could create a common component and import to both!

src/User/ProfilePage.jsx
```diff
@@ -78,6 +81,14 @@ const ProfilePage = ({
         <p className="occupation">{`Occupation: ${occupation}`}</p>
         {isCurrentUser && displayManager()}
       </StyledCard>
+      {isCurrentUser && (
+        <StyledCard className="employeesCard">
+          <StyledCardHeading className="employeeListHeader">
+            Employees:
+          </StyledCardHeading>
+          {employees.length > 0 && <EmployeeList employees={employees} />}
```
tonymaynard: Should this check be at a higher level? If you have 0 employees under your management it would still say `Employees:` Could potentially use a ternary operator to display something like `No linked employees`.

src/User/ProfilePage.test.jsx
```diff
@@ -40,8 +40,45 @@ describe("ProfilePage", () => {
...
+  ];
+
+  const employeeName = (employeeList, employeeIndex) => {
+    const employee = employeeList[employeeIndex];
+    return `${employee.firstName} ${employee.lastName}`;
+  };
+
+  const mockNoEmployees = [];
+
+  it("renders current user's profile information with manager and employees", async () => {
```
tonymaynard: Would probably treat the employees as a seperate unit test

src/User/EmployeeList.jsx
MerrettM: I think that is in another branch so when I resolve the merge conflicts in a moment I can do that.

src/User/ProfilePage.jsx
MerrettM: Good idea, I didn't want the card to disappear altogether so that would solve it.

src/User/ProfilePage.test.jsx
MerrettM: Will do!

src/User/EmployeeList.jsx
```diff
@@ -0,0 +1,36 @@
...
+`;
+const StyledEmployeesLink = styled(Link)`
+  display: inline-block;
+  text-decoration: none;
+  color: #7a517d;
+`;
+
+const EmployeeList = ({ employees }) => {
+  return (
+    <StyledEmployeesList className="employeeList">
```
cpn1306: Is the class "employeeList" used anywhere, do you need it?

MerrettM: yeah it's used in `ProfilePage` line 89

src/Navbar.test.js
```diff
@@ -41,28 +41,25 @@ describe("Navbar", () => {
...
     expect(wrapper.find(".register").length).toBe(0);
     expect(wrapper.find(".login").length).toBe(0);
-    expect(wrapper.find(".profilePage").get(0).props.children).toBe("Profile");
-    expect(wrapper.find(".logout").get(0).props.children).toBe("Logout");
-    expect(wrapper.find(".addConference").get(0).props.children).toBe(
-      "Add Conference"
-    );
+    expect(findElement(".profilePage")).toBe("Profile");
+    expect(findElement(".logout")).toBe("Logout");
+    expect(findElement(".addConference")).toBe("Add Conference");
```
longstaff-elsevier: not sure what this is doing, do you mean the text is "Add Conference"? why is the element equal to be a string?

src/User/EmployeeList.test.jsx
```diff
@@ -0,0 +1,88 @@
+import React from "react";
+import Adapter from "enzyme-adapter-react-16";
+import { configure } from "enzyme";
+import { Router } from "react-router-dom";
+import { createMemoryHistory } from "history";
+import { wrapper, setMountedWrapper } from "../TestUtils";
+import EmployeeList from "./EmployeeList";
+
+jest.mock("react-router-dom", () => {
```
longstaff-elsevier: Should mock before imports

src/Navbar.test.js
MerrettM: findElement is a method which simplifies `wrapper.find(".addConference").get(0).props.children`

PR #38, opened 2020-01-06
GP19-184: Unmocked API Request in ConferenceDetails Test

MerrettM:
Needs to be mocked so that running tests doesn't spit out huge amounts of error messages despite the tests passing.


MerrettM:
For context the error messages were due to the fact a console log was being fired due to an api request returning an error as it was in a test component and not a real one.

PR #39, opened 2020-01-07
Gp19 176 react testing library test for simple component

MerrettM:
This ticket is to practice using react-testing-library on a simple component, another ticket has been made for testing a more complex component to be done at a later date.


MerrettM:
I am aware that this test uses multiple different element selection methods, this is on purpose to serve as an example as this is the first react-testing-library test any of us have written.

src/Conference/ConferenceItem.jsx
```diff
@@ -41,11 +41,11 @@ const ConferenceItem = ({ conference }) => {
   return (
     <StyledConferenceItem>
       <StyledCard>
-        <StyledCardHeading>{name}</StyledCardHeading>
-        <p>{topic}</p>
+        <StyledCardHeading role="header">{name}</StyledCardHeading>
```
tonymaynard: Do you not mean role="heading"? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role

src/Conference/ConferenceItem.test.jsx
```diff
@@ -0,0 +1,31 @@
...
+test("displays conference item", async () => {
+  const mockConference = {
+    id: 1,
+    name: "Mock Conference",
+    topic: "Mock Topic",
+    dateTime: "2020-11-12T12:34:11Z",
+    city: "Mock City"
+  };
+
+  render(
```
tonymaynard: I'm not sure what screen is as I haven't used this library that much, but When I have screen is not needed. You can access `getByRole` etc from the render method. i.e. `const { getByText, getByRole } = render( <SomeThingToRender />)`

src/Conference/ConferenceItem.test.jsx
```diff
@@ -0,0 +1,31 @@
...
+    <MemoryRouter initialEntries={[{ pathname: "/", key: "testKey" }]}>
+      <ConferenceItem conference={mockConference} />
+    </MemoryRouter>
+  );
+
+  expect(screen.getByRole("header").textContent).toBe(mockConference.name);
+  expect(screen.getByTestId("topic").textContent).toBe(mockConference.topic);
+  expect(screen.getByText("/", { exact: false }).textContent).toBe(
```
tonymaynard: why dont you use a data attribute here?

src/Conference/ConferenceItem.test.jsx
MerrettM: Seems to be just another way of doing the same thing you just described as far as I can tell, I prefer your way so I've gone with that.

src/Conference/ConferenceItem.test.jsx
MerrettM: Are you referring to the `.toBe("12/11/2020")`? I'm using that because if I used the `dateTime` from `mockConference` I'd have to put it through the same custom `getDatestring(new Date(dateTime))` as the component does.

MerrettM: Although actually that function does have its own dedicated function so actually it might be okay to use it in the test?

tonymaynard: No sorry I meant `expect(getByTestid("date")` as its much more readible you can still use the getByText For the link

MerrettM: I'll make it `expect(getByTestid("date")`, but I'll leave the time one how it is as I wanted to demonstrate the `exact: false` functionality.

RaMorey:
That is alooot of package lock changes, did you update everything?

MerrettM:
> That is alooot of package lock changes, did you update everything?

I had to do a lot of `npm i` and `npm audit fix` stuff that might have done that?

PR #40, opened 2020-01-09
GP19-182: fav conference and employees conditionally rendered

GBurley-Jones:
I have not changed the size and shape of the cards as there is a separate ticket for that :)

### DIFF
--- a/src/User/ProfilePage.jsx
+++ b/src/User/ProfilePage.jsx
@@ -68,31 +68,33 @@ const ProfilePage = ({
   };
 
   const displayFavouriteConferences = () => {
-    if (favouriteConferences.length > 0) {
-      return (<>
-        <StyledCard profileCard>
-          <StyledCardHeading className="favouriteConferencesHeader">
-            Favourited Conferences:
-                </StyledCardHeading>
-          <FavouriteConferenceList
-            className="favouriteConferencesList"
-            favouriteConferences={favouriteConferences}
-          />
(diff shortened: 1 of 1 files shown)


src/User/ProfilePage.jsx
```diff
@@ -67,6 +67,35 @@ const ProfilePage = ({
     }
   };
 
+  const displayFavouriteConferences = () => {
+    if (favouriteConferences.length > 0) {
+      return (<>
```
RaMorey: It would be more readable if the fragment was on the next line

src/User/ProfilePage.jsx
```diff
@@ -67,6 +67,35 @@ const ProfilePage = ({
     }
   };
 
+  const displayFavouriteConferences = () => {
+    if (favouriteConferences.length > 0) {
```
RaMorey: I think you can change this to favouriteConferences.length instead of checking if it's greater than 0, as 0 is falsy

src/User/ProfilePage.jsx
```diff
@@ -67,6 +67,35 @@ const ProfilePage = ({
...
+      return (<>
+        <StyledCard profileCard>
+          <StyledCardHeading className="favouriteConferencesHeader">
+            Favourited Conferences:
+                </StyledCardHeading>
```
RaMorey: It's nice to close JSX Tags on the same indentation they were opened

